PR #343, opened 2022-06-21
Add diffing of before/after examples

isomorpheme:
Copied from https://github.com/haskell/error-messages/pull/180, see there for some previous discussion.


isomorpheme:
Not going to work on this in the near future, so closing.

PR #352, opened 2022-10-21
GHC-51179: fix typo in LambdaCase example

Prillan:
Trivial fix, but it gives me an opportunity to say that this is an excellent project and website!


david-christiansen (approved):
Nice catch!

david-christiansen:
And thank you for your help in making it better :-)

Issue #355, opened 2022-10-21
What's the lifecycle of an error code?


michaelpj:
I have a few questions about how change is managed for error codes and explanations. Lots of this is fuzzy, but I think these are practical questions for people using this!

1. What are the criteria for an error entering this index? It exists in a published version of a project which has applied to get some namespace allocated in the index?
2. What happens if an error is removed? Is the old code taken forever? Does the old explanation exist forever? Do we need some metadata about what versions an error exists in?
3. What happens if the message for an error is changed/improved? Should the explanation be updated to match the new text? If the change is significant, should we retain both versions? Or should a substantially changed error get a new code? Is there some rubric for what constitutes the "same" error?

david-christiansen:
>    What are the criteria for an error entering this index? It exists in a published version of a project which has applied to get some namespace allocated in the index?

Either that, or it has appeared in a version that will be imminently published with a very high degree of probability (that's the situation for GHC 9.6 at the moment)

> What happens if an error is removed? Is the old code taken forever? Does the old explanation exist forever? Do we need some metadata about what versions an error exists in?

The code is indeed taken forever, as is done with the Rust compiler. The old explanation will persist forever, though we may want to hide them from the main list by default if they haven't been in a released version for 10 years or so. I think time and experience will suggest good practices here. We already have metadata about the version in which the code was introduced, and we can easily add it for when codes are removed.

> What happens if the message for an error is changed/improved? Should the explanation be updated to match the new text? If the change is significant, should we retain both versions?

I think we should keep both wordings around if it makes the page easier to understand. Basically, this site is all about communicating to humans, so an added or removed comma or a typo fix probably doesn't require explicit calling out, while replacing "the skolem would escape" with "the type argument is not available here" would probably call for keeping both versions.

> Or should a substantially changed error get a new code? Is there some rubric for what constitutes the "same" error?

Here we have to fall back on the fact that it's documentation for people. The code is associated with the meaning, not the wording. A change in wording but not meaning keeps the same code, while a change in meaning means that the old error was retired and a new one was created. There's no formal rubric - this will be a matter for the good judgment of tool developers.

I suppose I should write this up more cleanly and add a "recommendations for tool developers" document. Thanks for pointing out the need!

david-christiansen:
I think that this is resolved by [ae5315f](https://github.com/haskellfoundation/error-message-index/commit/ae5315fc1e7fefe733b6749638cdf0f09f4e97b6) - please reopen if not!

PR #356, opened 2022-10-21
Add docs for tool developers

david-christiansen:
I hope that this addresses the questions from #355.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ Haskell tooling emits a variety of errors, warnings, and other messages. In the 
 
 Please see [CONTRIBUTING.md](./CONTRIBUTING.md) for instructions on improving the site. Thanks!
 
+If you're a Haskell tool developer who would like to integrate your project with the site, please refer to [tool-developers.md](./tool-developers.md) for recommendations regarding the error codes themselves.
+
 ## Running Locally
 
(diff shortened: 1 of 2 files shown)


michaelpj (approved):
This is very helpful, thanks!

Some more questions:

I think that this effectively means that _error code assignment_ and _explanation provision_ are coupled: in order to get an error code I have to write an explanation of it. Is there any way for me to get some codes at the cost of some "explanation debt"? Suppose I want to convert all my errors to have codes, but I don't have time to write all the explanations right now, it would be nice to be able to allocate all the codes, and then later add the explanations. Or maybe if I "own" the namespace I can just do this?

Relatedly, I think maybe it would help to have a few user workflows written out. E.g. here's what I think happens if you add a new error:
1. PR to project adding new error, assign error code according to your namespace's allocation strategy (?). No need to do anything to the error index (?).
2. PR is merged to project.
3. Add an explanation to the error index for the new code. Can do this at any time including before (?) or after (?) releasing the software with the new error.

david-christiansen:
I've uploaded a new version of the docs that I hope addresses @michaelpj's questions.

michaelpj (approved):
Great stuff!

david-christiansen:
Thanks for asking good questions, I'm glad to have the metadocs also improve :-)

Issue #347, opened 2022-10-20
Code examples are sometimes too wide for their box


david-christiansen:
Right now, code examples that are too wide overflow the allowed space. This is especially a problem on mobile browsers.

Screenshots:
![Screenshot from 2022-10-20 18-58-01](https://user-images.githubusercontent.com/115330/197012250-2866b8cb-3d35-4d2d-849d-8ee13b397de8.png)

![Screenshot_20221020-185859_Firefox](https://user-images.githubusercontent.com/115330/197012349-3d0c3f3b-b41a-4ffa-9aa1-be4af902ce7f.jpg)

klequis:
I'm happy to address this issue. I have received a lot of help from folks in the Haskell community and have been looking for a way to give something back. If you would like me to do so I can get back to you in a few days with my thoughts on changes?

david-christiansen:
It would be great if you could do that! Thanks!

klequis:
It seems like most sites, including my own, use a horizontal scrollbar when a code example is too wide for the screen.

For the error message, which isn't in a table, the overflow can be corrected with 
```css
pre {
  overflow-x: auto;
}
```
However, this doen't work in a table (atleast I couldn't get it to do so or find a answer searching) and therefore doesn't work for the before and after examples.

Personally, I think the before and after should be horizontal (side-by-side) on screens that are wide enough but should switch to a vertical layout for for narrow screens / phones. To achieve this the examples need to be taken out of a table and [flexbox](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox) is introduced to the CSS.

I created a live demo to show this http://haskell-error-index.s3-website-us-west-2.amazonaws.com/, but that's an odd looking link from a stranger so here are screenshots.

## Full-screen

![full-screen](https://user-images.githubusercontent.com/18337376/199619657-b92c1ba2-a479-4e1f-bcc8-3d92efdf69a4.png)

## iPad portrait

![ipad-portrait](https://user-images.githubusercontent.com/18337376/199619689-f08f4671-381e-42b4-a1c8-6c59eed62a42.png)

## Width of iPhone 11
*but made taller*

![width-of-iphone11](https://user-images.githubusercontent.com/18337376/199619748-83c543ba-f3c2-4c04-96ef-29028d610730.png)

## Extra narrow
*to force the before and after to overflow*

[...]

Issue #348, opened 2022-10-20
Message table too wide on mobile browsers


david-christiansen:
The table of messages on the front page of the generated site is too wide on mobile browsers:
![Screenshot_20221020-190136_Firefox](https://user-images.githubusercontent.com/115330/197012748-f47e633d-f245-46ac-a433-99ede8b05b49.jpg)

Perhaps the number and message fields can be stacked, and the "since" field omitted on very narrow screens?

klequis:
I looked around for different methods for making tables responsive. The short story is they aren't very responsive but there are a handfull of tricks out there to help.

The solution I see most often suggested is [Responsive Data Tables](https://css-tricks.com/responsive-data-tables/).

Here is a working mockup: http://haskell-error-index.s3-website-us-west-2.amazonaws.com/
I just did a quick tweak of the example without changing any of the formatting, so it doesn't look like a Haskell site and there is lots of wasted space. However, it does demonstrate the functionality.

I would like to suggest that the column labels (now kind of row labels) add little value in the small screen layout and think eliminating them in favor of having one column would improve readability and decrease vertical scroll. Let me know what you think.

All the code for the example is in these two files
- https://github.com/klequis/css-tricks-tbl/blob/main/public/index.html
- https://github.com/klequis/css-tricks-tbl/blob/main/public/styles.css

PR #357, opened 2022-11-11
fixes issue #347

klequis:
You can preview the changes at http://haskell-error-index.s3-website-us-west-2.amazonaws.com/. Unlike the previous site I had here, this is the actual site built with Hakyll.

There are two things you may object to. Normally I would not change code formatting. However, I have been sick for over a week (still am) and am now way behind on other things so don't have time to correct.
1. I reordered the css in default.css. There are fewer changes than it appears.
2. My editor automatically formats html differently. It does 2 spaces and some other indent differences. I was going to correct but then I noticed some of the html was already 2 spaces so I left it all like that.

In the event that issue #348 remains open, I expect to have time in early December to contribute more.


klequis:
Also fixes #296  which seems to be a duplicate.

david-christiansen:
Once CI passes, I'll give it a local test. The changes to the various files are more than I can figure out with my eyeballs.

Thanks!

david-christiansen:
OK, I've tested it all out. I ran into some issues with backgrounds of preformatted text being applied only to where there actually is text, rather than to the containing box, but I managed to fix that myself.

Thank you very much!

PR #363, opened 2022-11-29
Stack Error Message [S-4804]

47Cid:


david-christiansen:
Thank you very much for the contribution!

It looks to me like the example YAML files contain quite a bit of content, but that they could probably be made smaller to more directly illustrate the problem. In particular, I suspect that they don't need test suites, GHC options, and the like in order to trigger the message.

The reason for using files that are as small as possible to trigger an error in various ways is that it makes it easier for someone reading the documentation to visually pattern-match on a feature that they see in their own projects and thus recognize a solution.

Would you mind making them smaller, if possible?

Thanks!

message-index/messages/S-4804/does-not-match/before/package.yaml
```diff
@@ -0,0 +1,15 @@
+name:                blah
+
+dependencies:
+- base >= 4.7 && < 5
+- say >= 0.2.0.1
```
david-christiansen, line 5: What about picking a ludicrously large version number here, just so the example remains valid after the next release?
```suggestion
- say >= 100.2.0.1
```

message-index/messages/S-4804/does-not-match/index.md
```diff
@@ -0,0 +1,20 @@
...
+
+The specific version of the package is not provided by the snapshot. The error that results from `stack build` is:
+```text
+ Error: [S-4804]
+       Stack failed to construct a build plan.
+       
+       While constructing the build plan, Stack encountered the following exceptions:
+       
+       In the dependencies for blah-0.1.0.0:
+           say-0.1.0.1 from Stack configuration does not match >=0.2.0.1
```
david-christiansen, line 13:
```suggestion
           say-0.1.0.1 from Stack configuration does not match >=100.2.0.1 
```

david-christiansen (commented):
Thanks! I've got one other thing that I noticed now that the files are shorter. What do you think?

david-christiansen (approved):
Looks great! Will merge once CI goes green.

Thanks again!

PR #364, opened 2022-12-12
Document another case of GHC-53786

mrgutkun:
A case-expression inside a do-block whose branch expression is missing a `do` parses unintuitively, resulting in a confusing GHC-53786. It seems worthwhile to document  this explicitly.

### DIFF
--- a/message-index/messages/GHC-53786/case-expr-in-do-block/index.md
+++ b/message-index/messages/GHC-53786/case-expr-in-do-block/index.md
@@ -2,7 +2,9 @@
 title: Case expression in do-block
 ---
 
-The expression in one branch of the case expression is a do-block that's missing a `do`; the whole case-expression is interpreted as LHS of a bind in the top-level do-block. See also: https://gitlab.haskell.org/ghc/ghc/-/issues/984
+The expression in one branch of the case expression is a `do`-block that's missing a `do`. This triggers a corner case in the grammar of Haskell, and the `case`-expression is interpreted as the pattern part of a bind statement in the top-level `do`-block. 
+
+More details are available at https://gitlab.haskell.org/ghc/ghc/-/issues/984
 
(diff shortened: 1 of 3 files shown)


message-index/messages/GHC-53786/case-expr-in-do-block/index.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Case expression in do-block
+---
+
+The expression in one branch of the case expression is a do-block that's missing a `do`; the whole case-expression is interpreted as LHS of a bind in the top-level do-block. See also: https://gitlab.haskell.org/ghc/ghc/-/issues/984
```
david-christiansen, line 5:
```suggestion
The expression in one branch of the case expression is a `do`-block that's missing a `do`. This triggers a corner case in the grammar of Haskell, and the `case`-expression is interpreted as the pattern part of a bind statement in the top-level `do`-block. 

More details are available at https://gitlab.haskell.org/ghc/ghc/-/issues/984
```

david-christiansen: I think that these changes render it a bit easier to follow - the original version had a fairly long sentence that took me a couple of passes. What do you think?

david-christiansen (changes requested):
Nice catch!

I think that the exact phrasing of the description is a bit difficult to read. I suggested a rephrasing - did I capture the meaning correctly?

Thanks!

message-index/messages/GHC-53786/case-expr-in-do-block/index.md
mrgutkun: This does read better, thanks!

david-christiansen (approved):

david-christiansen:
Thanks again!

PR #373, opened 2023-01-14
Add guidance about GHCup's pre-release channel

tfausak:
I ran into this while trying to install ghc GHC 9.6.1-alpha1 using GHCup. The GHCup-000100 page already contains enough information to piece this together, but I figured it was worth an explicit callout.

### DIFF
--- a/message-index/messages/GHCup-00010/index.md
+++ b/message-index/messages/GHCup-00010/index.md
@@ -64,4 +64,4 @@ Other reasons you might get this error:
 
 * The version of the tool you specified doesn't exist at all, e.g. `ghcup install ghc blah`.
 * The version of the tool you specified wasn't added to GHCup yet, but will likely be added soon. Please [raise an issue here](https://github.com/haskell/ghcup-metadata/issues) and make sure to update ghcup's internal list of available tools with `ghcup upgrade`.
-
+* The version of the tool you specified was a pre-release, like an alpha or release candidate. Add [a release channel](https://www.haskell.org/ghcup/guide/#pre-release-channels) to make it available. For example: `ghcup config add-release-channel https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.7.yaml`.


message-index/messages/GHCup-00010/index.md
```diff
@@ -64,4 +64,4 @@ Other reasons you might get this error:
 
 * The version of the tool you specified doesn't exist at all, e.g. `ghcup install ghc blah`.
 * The version of the tool you specified wasn't added to GHCup yet, but will likely be added soon. Please [raise an issue here](https://github.com/haskell/ghcup-metadata/issues) and make sure to update ghcup's internal list of available tools with `ghcup upgrade`.
-
+* The version of the tool you specified was a pre-release, like an alpha or release candidate. Add [a release channel](https://www.haskell.org/ghcup/guide/#pre-release-channels) to make it available. For example: `ghcup config add-release-channel https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.7.yaml`.
```
david-christiansen, line 67: Where does `https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.7.yaml` come from? The reason I ask is that people who are skimming for solutions might miss the "for example" and then get stuck. A link to a page that discusses how to select this file would be really useful here.

david-christiansen (changes requested):

david-christiansen:
Thanks!

message-index/messages/GHCup-00010/index.md
tfausak: It comes from the documentation page linked in the previous sentence: https://www.haskell.org/ghcup/guide/#pre-release-channels

david-christiansen: What I don't see at the link is where to get prerelease channels from, but I suppose that's more an issue with the linked docs than with this PR. Thanks!

tfausak: Oh, I see. As far as I know, there's no list of pre-release channels.

PR #380, opened 2023-01-26
Update index.html

hsyl20:
- mention only GHC as an example since some `ghcup` errors are indexed too
- soften the wording about GHC returning codes for its messages (it's still a work in progress)

### DIFF
--- a/message-index/index.html
+++ b/message-index/index.html
@@ -7,7 +7,7 @@ title: The Haskell Error Index
 <p>
   This site describes the various messages that can be returned by Haskell-related tools, including both errors and warnings.
   For example GHC, the most-commonly-used Haskell implementation, started emiting a code with the format <code>[GHC-12345]</code> for some of its
-  messages since version 9.6.1.
+  messages in version 9.6.1.
   These codes can be looked up below for further information.
 </p>
 
--- a/message-index/index.html
+++ b/message-index/index.html
@@ -6,7 +6,7 @@ title: The Haskell Error Index
 
 <p>
   This site describes the various messages that can be returned by Haskell-related tools, including both errors and warnings.
-  For example GHC, the most-commonly-used Haskell implementation, started emiting a code with the format <code>[GHC-12345]</code> for some of its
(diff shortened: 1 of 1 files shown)


message-index/index.html
```diff
@@ -5,10 +5,10 @@
 <h2>Welcome</h2>
 
 <p>
-  This site describes the various messages that can be returned by
-  GHC, the most-commonly-used Haskell implementation, including both errors and warnings. Beginning with version
-  9.6.1, GHC emits a code with the format <code>[GHC-12345]</code> for each
-  message. These codes can be looked up below for further information.
+  This site describes the various messages that can be returned by Haskell-related tools, including both errors and warnings.
+  For example GHC, the most-commonly-used Haskell implementation, started emiting a code with the format <code>[GHC-12345]</code> for some of its
+  messages since version 9.6.1.
```
david-christiansen, line 10:
```suggestion
  messages in version 9.6.1.
```
To match "started". Alternatively, it could read "GHC has emitted a code with the format ... since version 9.6.1."

message-index/index.html
```diff
@@ -5,10 +5,10 @@
 <h2>Welcome</h2>
 
 <p>
-  This site describes the various messages that can be returned by
-  GHC, the most-commonly-used Haskell implementation, including both errors and warnings. Beginning with version
-  9.6.1, GHC emits a code with the format <code>[GHC-12345]</code> for each
-  message. These codes can be looked up below for further information.
+  This site describes the various messages that can be returned by Haskell-related tools, including both errors and warnings.
+  For example GHC, the most-commonly-used Haskell implementation, started emiting a code with the format <code>[GHC-12345]</code> for some of its
```
david-christiansen, line 9:
```suggestion
  For example GHC, the most-commonly-used Haskell implementation, started emiting a code with the format <code>[GHC-12345]</code> for its
```
I think the "some" is needlessly pessimistic - the messages that most people encounter on a regular basis are covered, after all. I agree that "for each" is not technically correct, but I fear that the technically correct version is an overcorrection, and that strategic vagueness better communicates the state of the world.

hsyl20:
Suggestions applied, thanks @david-christiansen!

david-christiansen (approved):

PR #381, opened 2023-02-01
Improve some GHCup errors

hasufell:


message-index/messages/GHCup-00160/index.md
```diff
@@ -13,6 +13,7 @@ Potential underlying causes for this errors are:
 
 * No internet connection
 * Unreliable disk or Internet leading to a corrupted file
+* broken curl or wget (e.g. installed through [snap](https://snapcraft.io/))
```
david-christiansen, line 16:
```suggestion
* Non-functional `curl` or `wget` (users have reported difficulties when these tools are installed through [snap](https://snapcraft.io/))
```

I'd rather be specific than just throw shade

message-index/messages/GHCup-00160/index.md
```diff
@@ -32,22 +33,31 @@ $ ghcup list
...
 Alternatively try with wget:
 
 ```
+cd "$(ghcup whereis cachedir)"
 wget -O- https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.7.yaml
 ```
 
-If the download succeeds, but the error persists, try examining the file:
+If the download fails, your curl/wget installation might be broken. On linux, make sure you've installed
+these from your package manager and not via [snap](https://snapcraft.io/) or similar. On windows, make
```
david-christiansen, line 54:
```suggestion
these from your package manager and not via [snap](https://snapcraft.io/) or similar. On Windows, make
```

What counts as "similar"? How would users know?

message-index/messages/GHCup-00160/index.md
```diff
@@ -32,22 +33,31 @@ $ ghcup list
...
+sure you use `curl.exe` and not `curl` (which is a cmdlet) when testing. It should be pre-installed.
+
+If the download succeeds, but the error persists, try examining the file and verify it's valid YAML:
```
david-christiansen, line 57:
```suggestion
If the download succeeds, but the error persists, try examining the file and verify that it's valid YAML:
```

message-index/messages/GHCup-05841/index.md
```diff
@@ -21,6 +21,8 @@ $ ghcup install ghc -u https://nothing/ghc.tar.gz 9.4.3
 ```
 
 This error is a combination of [GHCup-00841](https://errors.haskell.org/messages/GHCup-00841/index.html) and `GHCup-05000`.
+Sometimes you may encounter [GHCup-00160](https://errors.haskell.org/messages/GHCup-00160/index.html) instead, depending
+on how you invoked `ghcup`.
 
 Reasons for this errors:
```
david-christiansen, line 27:
```suggestion
Reasons for this error:
```

That wasn't a change in this PR, but I noticed it :-)

david-christiansen (changes requested):
First off, sorry for the late answer here.

I suggested a few rephrasings that I think make the text both more specific and less likely to be hurtful. There were also a few capitalization fixes.

message-index/messages/GHCup-00160/index.md
hasufell: > What counts as "similar"? How would users know?

Flatpak, AppImage and all the other cr... err, creative solutions.

david-christiansen: Those are common ways to distribute software that are generally recommended by distributions these days. If people have reported difficulties when using them, then we should definitely document those difficulties, but we should expect that people will be using them more and more in the future. I don't think that insulting them is going to make any positive changes here.

So the question remains - are users reporting difficulties using `curl` and/or `wget` from Flatpak or AppImage? If so, we should list them, and if not, then I suggest we just list Snap as something to avoid for now.

message-index/messages/GHCup-00160/index.md
```diff
@@ -32,22 +33,31 @@ $ ghcup list
...
 
 Alternatively try with wget:
 
 ```
+cd "$(ghcup whereis cachedir)"
 wget -O- https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.7.yaml
 ```
 
-If the download succeeds, but the error persists, try examining the file:
+If the download fails, your curl/wget installation might be broken. On linux, make sure you've installed
```
david-christiansen, line 53:
```suggestion
If the download fails, your `curl`/`wget` installation might be broken. On Linux, make sure you've installed
```

message-index/messages/GHCup-00160/index.md
```diff
@@ -32,22 +33,31 @@ $ ghcup list
...
+these from your package manager and not via [snap](https://snapcraft.io/) or similar. On Windows, make
```
david-christiansen, line 54:
```suggestion
these from your package manager and not via [snap](https://snapcraft.io/), which has been reported to cause problems with `ghcup`. On Windows, make
```

david-christiansen (changes requested):

message-index/messages/GHCup-00160/index.md
hasufell: > Those are common ways to distribute software that are generally recommended by distributions these days

I don't think that's generally true. The distributions that I've worked with do not seem to recommend those ways (e.g. gentoo, exherbo, archlinux).

And I also generally disagree that those are reasonable architectures.

hasufell: > If so, we should list them, and if not, then I suggest we just list Snap as something to avoid for now.

My recommendation is and will remain to install curl/wget via your package manager and *nothing* else. I will make this more clear.

hasufell: I think I have sufficient autonomy to make that call for GHCup, which makes use of curl and wget.

david-christiansen: I think that:

> My recommendation is and will remain to install curl/wget via your package manager and nothing else.

is a fine way to write it.

My concerns here are that the documentation site is not provoking flame wars or causing others to feel insulted and that any recommendations given are specific and actionable.

message-index/messages/GHCup-00160/index.md
```diff
@@ -32,22 +33,31 @@ $ ghcup list
...
 Alternatively try with wget:
 
 ```
+cd "$(ghcup whereis cachedir)"
 wget -O- https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.7.yaml
 ```
 
-If the download succeeds, but the error persists, try examining the file:
+If the download fails, your `curl`/`wget` installation might be broken. On Linux, make sure you've installed
+these from your package manager and not via [snap](https://snapcraft.io/), [Flatpak](https://flatpak.org/) or [AppImage](https://appimage.org/). GHCup does neither recommend nor support those tools.
```
david-christiansen, line 54:
```suggestion
these from your package manager and not via [snap](https://snapcraft.io/), [Flatpak](https://flatpak.org/) or [AppImage](https://appimage.org/). GHCup neither recommends nor supports those tools.
```

message-index/messages/GHCup-00160/index.md
hasufell: > My concerns here are that the documentation site is not provoking flame wars or causing others to feel insulted and that any recommendations given are specific and actionable.

I totally agree and I'm a bit confused you feel the need to mention this.

I was simply signalling to you that I have a **very low opinion** on the mentioned technologies, because we're discussing this informally here, aren't we?

The documentation should indeed just explain what we're willing to support and what we're considering problematic without making blanket statements.

I think that the current wording achieves that.

david-christiansen: I think that we've been caught out here by the lack of bandwidth of text communication here.

I mentioned those principles out of what is probably an overabundance of clarity: I wanted to make it clear that I was working based on those, rather than on a preference for or against some particular software distribution method. I like being explicit about background assumptions where possible, but explicitly communicating obvious-seeming principles certainly runs the risk of implying that someone else is behaving badly.

I'll commit the grammar fix and merge if it's OK with you. Thanks!

hasufell: Sure

hasufell:
Thanks

Issue #350, opened 2022-10-20
No convenient programmatic access to the search


mjdominus-ly:
I want to set up a search macro for my browser's URL bar, so I can type `err blah` and have it go visit `https://errors.haskell.org/?q=blah`.

There doesn't seem to be any way to do that, because the search bar is an embedded JS thing with no API.

Ultimately there ought to be a real API where a program can post a string or error code and get back a description or a URL, but that would be a separate bug report.  All I am currently requesting is a distinct URL for each search.

mjdominus-ly:
The Javascript can look at the query parameters and, if there are any,
prepopulate the input field with the query parameters instead of with the
placeholder text.

david-christiansen:
Right - would this be sufficient for your use case?

mjdominus-ly:
Yes, I think so.

david-christiansen:
Great to hear!

hasufell:
I have a similar problem, which might have a separate solution. I want a link to ALL GHCup errors. That doesn't have to be done with the filter, we could just have a subpage for grouping.

BinderDavid:
This has been fixed by #387 If there is still some missing feature please shout :)

Issue #378, opened 2023-01-19
Allow to link to all errors of a specific tool


hasufell:
Follow-up of https://github.com/haskellfoundation/error-message-index/issues/350#issuecomment-1382672906

I'd like to have a link such as https://errors.haskell.org/messages/GHCup that shows all and only GHCup errors.

david-christiansen:
This would be nice!

BinderDavid:
@hasufell The site is now online at https://errors.haskell.org/ghcup/ Thanks to @haglobah who implemented this feature :)

PR #387, opened 2023-03-04
Implement setting query from URL

TheBFL:
Implements the feature requested in issue #350. The site now auto-fills the contents of the search box with whatever is given in the query parameter `?q=...`, if it is given.


david-christiansen:
This works great - thanks!

I don't think it's a complete solution to #378, as that would also ideally work without JS, but it may solve some use cases. Will merge once CI goes green.

PR #391, opened 2023-03-07
Document GHC-01239 (closes #288)

### LINKED ISSUE #288: PsErrIfInFunAppExpr -> 1239

jhrcek:
Let me try my hand on documenting one of these error messages.

### DIFF
--- a/message-index/messages/GHC-01239/example1/after/IfInFunAppExpr.hs
+++ b/message-index/messages/GHC-01239/example1/after/IfInFunAppExpr.hs
@@ -1,0 +1,14 @@
+{-# LANGUAGE BlockArguments #-}
+module IfInFunAppExpr where
+
+-- Requires BlockArguments language extension to work
+example :: Int
+example = length if True then [1] else
+
+-- Works without BlockArguments
+example :: Int
+example = length (if True then [1] else [])
+
+-- Works without BlockArguments
+example2 :: Int
+example2 = length $ if True then [1] else []
--- a/message-index/messages/GHC-01239/example1/after/IfInFunAppExpr.hs
+++ b/message-index/messages/GHC-01239/example1/after/IfInFunAppExpr.hs
@@ -3,12 +3,12 @@ module IfInFunAppExpr where
 
 -- Requires BlockArguments language extension to work
 example :: Int
-example = length if True then [1] else
-
--- Works without BlockArguments
-example :: Int
-example = length (if True then [1] else [])
+example = length if True then [1] else []
 
 -- Works without BlockArguments
 example2 :: Int
-example2 = length $ if True then [1] else []
+example2 = length (if True then [1] else [])
+
+-- Works without BlockArguments
+example3 :: Int
+example3 = length $ if True then [1] else []
--- a/message-index/messages/GHC-01239/example1/index.md
+++ b/message-index/messages/GHC-01239/example1/index.md
@@ -3,7 +3,7 @@ title: Unexpected if expression in function application
 ---
 
 To pass If-Then-Else expressions as function arguments we either have to surround them in parentheses,
-use function application operator `($)` or enable [`BlockArguments`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/block_arguments.html#extension-BlockArguments) extension.
+use the function application operator `($)` or enable the [`BlockArguments`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/block_arguments.html#extension-BlockArguments) extension.
 
 ## Error Message
 ```
--- a/message-index/messages/GHC-01239/example1/index.md
+++ b/message-index/messages/GHC-01239/example1/index.md
@@ -1,0 +1,19 @@
+---
+title: Unexpected if expression in function application
+---
+
+To pass if-then-else expressions as function argument we either have to surround it in parentheses,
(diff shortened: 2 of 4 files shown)


message-index/messages/GHC-01239/index.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Unexpected if expression in function application
+summary: If expression used as function argument
+severity: error
+introduced: TODO
```
jhrcek, line 5: I have no idea how to find out in which version of GHC this error was introduced. Any suggestions?

david-christiansen: All error codes are introduced in 9.6.1, so thus far, it's not so difficult :-)

We should make it more clear that this documents which release the _code_ was introduced in.

message-index/messages/GHC-01239/example1/after/IfInFunAppExpr.hs
```diff
@@ -0,0 +1,14 @@
+{-# LANGUAGE BlockArguments #-}
+module IfInFunAppExpr where
+
+-- Requires BlockArguments language extension to work
+example :: Int
+example = length if True then [1] else
```
david-christiansen, line 6:
```suggestion
example = length if True then [1] else []
```
Isn't this missing?

message-index/messages/GHC-01239/example1/index.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Unexpected if expression in function application
+---
+
+To pass If-Then-Else expressions as function argument we either have to surround it in parentheses,
```
david-christiansen, line 5:
```suggestion
To pass If-Then-Else expressions as function arguments, they must either be surrounded in parentheses,
```

message-index/messages/GHC-01239/example1/index.md
```diff
@@ -0,0 +1,19 @@
...
+use function application operator `($)` or enable [`BlockArguments`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/block_arguments.html#extension-BlockArguments) extension.
```
david-christiansen, line 6:
```suggestion
used after the function application operator `($)`, or be used where the [`BlockArguments`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/block_arguments.html#extension-BlockArguments) extension is enabled.
```

message-index/messages/GHC-01239/example1/index.md
david-christiansen: Or if my suggested rewrite doesn't fit your taste, at least:
```suggestion
To pass If-Then-Else expressions as function arguments, we either have to surround them in parentheses,
```

message-index/messages/GHC-01239/index.md
```diff
@@ -0,0 +1,19 @@
...
+```haskell
+ageMessage :: Int -> String 
+ageMessage age = if age < 18 then "You are too young to enter" else "Welcome to the club"
+
+putStrLn (ageMessage 10) -- You are too young to enter
+putStrLn (ageMessage 20) -- Welcome to the club
+```
+
+Thinking of If-Then-Else expression as a value, we might want to pass it as an input to a function.
+If you do that you can run into the following error.
```
david-christiansen, line 19: Using both "we" and "you" to refer to the reader is potentially confusing. What about:

```suggestion
Because If-Then-Else expressions return values, it makes sense to pass them as input to a function.
However, without language extensions, Haskell's grammar requires parentheses around them in function argument positions.
```

david-christiansen (changes requested):
I think there's a correctness issue in the example.

I also added a few comments about the text style.

david-christiansen:
And thank you!

message-index/messages/GHC-01239/example1/after/IfInFunAppExpr.hs
jhrcek: How did that creep in? :smile:

message-index/messages/GHC-01239/example1/index.md
jhrcek: Went with the first one as I find it more natural.

jhrcek:
Thanks for suggestions! Applied.

message-index/messages/GHC-01239/example1/index.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Unexpected if expression in function application
+---
+
+To pass If-Then-Else expressions as function arguments we either have to surround them in parentheses,
+use function application operator `($)` or enable [`BlockArguments`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/block_arguments.html#extension-BlockArguments) extension.
```
david-christiansen, line 6: I think the grammar is still a bit off. Here's a minimal fix:
```suggestion
use the function application operator `($)` or enable the [`BlockArguments`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/block_arguments.html#extension-BlockArguments) extension.
```

jhrcek: Thanks for pointing that out. I'm not native speaker and I always had a bit of an issue with articles :blush:

david-christiansen (approved):

david-christiansen:
Sorry for the slow turnaround here! And thank you for the contribution.

Issue #385, opened 2023-02-28
CSS dark theme for HLJS


david-christiansen:
Once the imminent merge of #383 is completed, we'll have a nice dark mode. However, the HLJS theme (used for styling code samples) should be either built to match the Hackage colors that we use or selected to look good in the dark. Right now, it's a bit of a stop-gap.

It would be great to:
 * [ ] Design/find a code-highlighting theme (perhaps we can share this with haskell.org and Haddock?) for light mode rather than getting one off the shelf
 * [ ] Design or find a theme for dark mode that also matches other Haskell web branding
 * [ ] Arrange for it to be swapped in the CSS along with the other colors

PR #394, opened 2023-03-24
Colors

david-christiansen:
This includes #393 for reasons of me being able to test the code right now :-)

Before:
Screenshot from 2023-03-24 13-41-26

After:
Screenshot from 2023-03-24 13-41-46

CSS variables have been used to capture the various syntactic categories from hljs.


david-christiansen:
Fixes #385

PR #396, opened 2023-03-29
Add error message for error GHC-44432

BinderDavid:
Encountered the error message and felt the itch to upstream the documentation :)


message-index/messages/GHC-44432/example1/index.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Type signature lacks an accompanying binding
+---
+
+If a toplevel type signature is given, then there also needs to be an accompanying binding.
```
david-christiansen, line 5: I'm not sure that readers will necessarily know what a "binding" is - that's specialist programming language researcher lingo.

What about instead saying "If a top-level type signature is given for a name, then the name must also be defined"?

david-christiansen (changes requested):
Thanks!

I have a concern about the term "binding", which I think should either be explained or used in a way that's obviously synonymous with a more ordinary word.

It may also be good to include an example with a more complicated pattern, like:
```
a, b :: Int
(a, b) = (2, 3)
```

message-index/messages/GHC-44432/example1/index.md
```diff
@@ -0,0 +1,18 @@
...
+
+In this example, a type signature was specified for `someBoolean`, but no binding was given.
+This error can be fixed by adding a binding which accompanies the type signature.
```
jhrcek, line 8: I have a practical suggestion (though nothing that should block this PR if you don't like it).

In my daily Haskell programming it almost never happens that I would just write type signature and forget to add accompanying binding. This error happens most ofthen when you're copying things around or renaming things and you make a typo in function definition or in type declaration, so they get out of sync this way.

Maybe that case could be metioned as a cause for this error too?

david-christiansen: A typo example would be great.
```
factorial :: Natural -> Natural
fatcorial n
  | n == 0 = 1
  | otherwise = n * factorial (n - 1)
```

message-index/messages/GHC-44432/example1/index.md
BinderDavid: Thanks, I just figured out that the same message is also emitted for bindings which are not on the toplevel. I will also add some examples for that case.

message-index/messages/GHC-44432/example1/index.md
BinderDavid: It turns out that GHC already suggests the right thing ^^

```
Main.hs:5:1: error: [GHC-44432]
    The type signature for ‘factorial’ lacks an accompanying binding
    Suggested fix: Perhaps use ‘fatcorial’ (Defined at Main.hs:6:1)
  |
5 | factorial :: Natural -> Natural
  | ^^^^^^^^^
```

message-index/messages/GHC-44432/example1/index.md
BinderDavid: Ok, I have removed the word "binding" from the explanations. I still use it in the title which exactly reproduces the short message emitted by GHC.

message-index/messages/GHC-44432/example1/index.md
BinderDavid: > 



> I have a practical suggestion (though nothing that should block this PR if you don't like it).
> 
> In my daily Haskell programming it almost never happens that I would just write type signature and forget to add accompanying binding. This error happens most often when you're copying things around or renaming things and you make a typo in function definition or in type declaration, so they get out of sync this way.
> 
> Maybe that case could be mentioned as a potential cause for this error too?

Thanks, I have added an example which arises due to a typo.

david-christiansen (approved):

PR #400, opened 2023-04-17
Run Ormolu in separate GitHub Action

BinderDavid:
This factors out the ormolu action which checks whether `site.hs` is formatted into its own action, which is only triggered when the file `site.hs` is changed.

I'm trying to improve the efficiency of the github actions, and this is the first step :)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -35,12 +35,6 @@ jobs:
     - name: Update the Cabal index
       run: cabal update
 
-    - name: Check if site.hs is formatted using ormolu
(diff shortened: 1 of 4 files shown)


BinderDavid:
Yes, the deploy github action would now run through even if `site.hs` is not formatted according to Ormolu. But I don't think it is a problem. The `main` branch is protected, so it is not possible to push directly to `main`, and all changes have to go through PRs where the Ormolu action is run. Even if we would somehow get into a situation where non-well-formatted code gets into `main`, this is not a very serious problem, only a cosmetic one, which could quickly get fixed.

david-christiansen (approved):

PR #401, opened 2023-04-17
Merge deploy actions

BinderDavid:
Removes the code duplication between `deploy.yml` and `deploy2.yml` by introducing a build matrix for the targets "githubpages" and "haskellorg".

@david-christiansen Do we still use the githubpages target at all, or is everything hosted by haskell.org?

This PR builds on top of #400


david-christiansen:
We're still on GitHub pages - the haskell.org infra migration is in step 1 of the following process:
1. Deploy to both and make sure both work
2. Set up aggregate log analysis to see which error 404s are hit most
3. Point DNS at haskell.org infra
4. Get rid of GH pages

david-christiansen:
Not sure why 4ba548888fcf73fa3c416b500ef0507c5a552610 is still showing up here after merging #400.

david-christiansen:
Given that this is part of a migration, I'm not so sure that we need the nicely-factored code - after all, it'll go down to just one version soon, and then it'll be easier to just delete a YAML file than to re-edit and remove the matrix.

How does this help?

BinderDavid:
>Not sure why https://github.com/haskellfoundation/error-message-index/commit/4ba548888fcf73fa3c416b500ef0507c5a552610 is still showing up here after merging https://github.com/haskellfoundation/error-message-index/pull/400.


I guess it's because merging the PR has added an additional merge commit to main? I can rebase the branch on main, if you want, but then we would have to wait for CI to finish again 🤷 

> Given that this is part of a migration, I'm not so sure that we need the nicely-factored code - after all, it'll go down to just one version soon, and then it'll be easier to just delete a YAML file than to re-edit and remove the matrix.
> How does this help?


The caching logic in all 3 files which are building the hakyll site is incorrect right now.  I will have to modify the build instructions for all files in order to get caching to work. And its easier to check that I have correctly made the modifications in 2 different files compared to three different files.

david-christiansen (approved):

david-christiansen:
This all makes sense - thanks!

PR #403, opened 2023-04-17
Fix github action cache

### LINKED ISSUE #402: Cache in Github Actions doesn't work
The cache in the github actions for both the deploy actions and the actions which are run on every PR doesn't work correctly. As a consequence, hakyll is rebuilt from scratch every time which can take half an hour. 

I think I know what the problem is: The key for the cache consists of the cabal file. I.e. we have the following line:
```yaml
key: cabal-cache-0-${{ hashFiles('message-index/message-index.cabal') }}
```
This means that the last time the cache was updated was the last time we changed the cabal file.
Since we invoke `cabal update` directly afterwards, we invalidate the cached content and build from scratch.

Proposed solution: Use a cabal.freeze file.

BinderDavid:
Fixes #402 

This PR fixes the cache problem for the PR CI action, but doesn't change the code for the deploy actions yet.
The reason is that for the deploy actions I have to also fix the relative paths (to the subdirectory `message-index`) if I want to use the `default: subdirectory: message-index` trick, and I don't want to mess with that in this PR.

Builds on top of #401

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,11 @@ If you're a Haskell tool developer who would like to integrate your project with
 ## Running Locally
 
 The site is built with the [Hakyll](https://jaspervdj.be/hakyll/) static site generator. To view the site locally, enter the `message-index` directory and run:
-```
+```console
 $ cabal run -- site watch
 ```
 or
-```
+```console
 $ stack run -- site watch
 ```
 which fires up an HTTP server on `localhost:8000`.


BinderDavid:
Ok, it seems to work. Time is now down to ~2min instead of 30-40min.

david-christiansen (approved):
I was initially skeptical of the `plan.json` approach because it seems to be an undocumented side effect of `cabal`. OTOH, Agda does this, and the worst case is that the caching action breaks (it's broken today). CI config is an artifact of its time and circumstances, not Great Software For The Ages.

PR #405, opened 2023-04-18
Add code action which runs HLint on site.hs

BinderDavid:


david-christiansen:
My experience in the past with HLint is that it contains a number of opinionated style bits that are on by default, but are certainly not uncontroversial improvements (like wanting to replace `A $ B $ C` with `A . B $ C`, or making things more point-free). This meant that I've always had to spend a bit of time configuring HLint before using it.

Does this GitHub action use a premade config that's less opinionated? I can't see its output in the Actions runs for this PR, unfortunately.

BinderDavid:
> like wanting to replace A $ B $ C with A . B $ C, or making things more point-free

That particular lint is in the `dollar` configuration group of HLint and should be disabled by default. Currently   `site.hs` is free of any hlint warnings, but I have just added a default `.hlint.yml` configuration file which can easily be extended to ignore lints we don't like.

>  I can't see its output in the Actions runs for this PR, unfortunately.
The action is configured to only trigger on changes to `site.hs`, and to only lint that file.
By default, hlint recursively searches for `.hs` files and lints them, but this leads to tons of linter warnings for all the various examples in the error-messages.

david-christiansen:
Thanks, this makes sense!

PR #406, opened 2023-04-18
Update the caching logic for the deploy action. 

BinderDavid:
The deploy action should use the same caching logic as pullrequest-ci.yml, since the content of the caches is not shared among different branches, only the caches of the default branch (main) is shared with the CI actions run on the PR branches


david-christiansen (approved):

david-christiansen:
Thank you again for putting so much work into this!

PR #407, opened 2023-04-19
Make 404 friendlier

david-christiansen:
Update the 404 page to use some [undocumented features](https://github.com/isaacs/github/issues/1527) of the GitHub "create page" feature to automatically provide the right filename and a template.

It seems that the "create page" query strings have worked this way for at least a couple of years now, and it'll be easy enough to tweak the page again if it changes.


david-christiansen:
I just had a chat in person with Sam D., and he's on board with this update as being sufficient to address his concerns right now.

BinderDavid (approved):

PR #408, opened 2023-04-19
Small tool for generating the scaffolding of a new error message.

BinderDavid:
Small tool for creating new error messages more easily.
```console
$> runghc helper-tool.hs 
This tool helps you create the scaffolding for a new error message on the error-message-index.
You can leave any of the text fields blank and fill them in by hand later.

For which system do you want to generate a message?
 1) GHC
 2) GHCup
 3) Stack
Input (Default = GHC): 1

What is the errorcode that you want to document.
Input: 123456

What is the title of the error message.
Input: Missing Binding

Give a short summary of the error message.
Input: A type signature was given, but the binding is missing

What is the severity of the diagnostic.
 1) Error
 2) Warning
 3) N/A
Input (Default = N/A): 1


For which version was this error message emitted for the first time?
Note: For GHC this is most likely 9.6.1.
Input: 9.6.1

How many examples should be generated?
Input: 2
Creating scaffolding for the following configuration:
MkTemplate {system = GHC, code = "123456", title = "Missing Binding", summary = "A type signature was given, but the binding is missing", severity = Error, warningflag = Nothing, introduced = "9.6.1", examples = 2}

```


message-index/helper-tool.hs
```diff
@@ -0,0 +1,216 @@
...
+
+readSystem :: IO System
+readSystem = do
+    putStrLn "For which system do you want to generate a message?"
+    putStrLn " 1) GHC"
+    putStrLn " 2) GHCup"
+    putStrLn " 3) Stack"
+    putStr "Input (Default = GHC): "
+    ln <- getLine
+    case ln of
```
david-christiansen, line 24: I'd compare these strings case-insensitively and modulo leading and trailing whitespace for the sake of friendliness

message-index/helper-tool.hs
```diff
@@ -0,0 +1,216 @@
...
+    putStr "Input: "
+    getLine
+
+-- Examples
+type Examples = Int
+
+-- | Only ask for examples if the system is GHC.
+readExamples :: System -> IO Examples
+readExamples GHC = do
+    putStrLn "How many examples should be generated?"
```
david-christiansen, line 114: What about interactively prompting for example names instead? Or, alternatively, have a command-line app for adding an example to an existing message?

The concern here is that we  end up with a ton of `example2` directories instead of things with meaningful names. Not that it's the end of the world, though.

david-christiansen:
This is a good idea! It'll simplify a lot of things. We have a shell script somewhere in the repo that does something similar, but this one seems much easier to use.

message-index/helper-tool.hs
BinderDavid: I prompt the user for names of the examples now.

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+    strip = f . f
+    f = reverse . dropWhile isSpace
+
+-- Querying for the system: GHC / GHCup / Stack
+
+data System = GHC | GHCup | Stack deriving (Show)
+
+readSystem :: IO System
+readSystem = do
+  putStrLn "For which system do you want to generate a message?"
```
david-christiansen, line 26: I think that we call them "tools" elsewhere in the docs, rather than "tools".

```suggestion
  putStrLn "For which tool do you want to generate a message?"
```
or even
```suggestion
  putStrLn "Which tool's error code do you want to document?"
```

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+  putStr "Input (Default = GHC): "
+  ln <- getLine
+  case normalize ln of
+    "1" -> pure GHC
+    "ghc" -> pure GHC
+    "2" -> pure GHCup
+    "ghcup" -> pure GHCup
+    "3" -> pure Stack
+    "stack" -> pure Stack
+    _ -> pure GHC
```
david-christiansen, line 39: Doesn't this mean that typing `stcak` would get me docs for GHC? I'd expect the default to happen for empty input, not necessarily garbage input.

david-christiansen:
```suggestion
    "" -> pure GHC
    _ -> do
      putStrLn "Didn't understand input. Please type a tool name or a number."
      readSystem
```

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+
+-- | We need to encode the error code as a string in order
+--   to preserve leading 0's.
+type ErrorCode = String
+
+readCode :: IO ErrorCode
+readCode = do
+  putStrLn "What is the numeric code that you want to document."
+  putStrLn "For example, enter \"01234\" if you want to document GHC-01234."
+  putStr "Input: "
```
david-christiansen, line 51: This may not work reliably without turning off line buffering on stdout.

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+      putStrLn "Could not parse the input as an integer. Only enter the numeric part of the error."
+      readCode
+    Just _ -> pure ln
+
+-- Title
+type Title = String
+
+readTitle :: IO Title
+readTitle = do
+  putStrLn "What is the title of the error message."
```
david-christiansen, line 64:
```suggestion
  putStrLn "What is the title of the error message?"
```

david-christiansen: Perhaps also include a quick description of what this is used for, e.g. "This is used as the title of the documentation page as well as in links to the page."

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+  putStr "Input: "
+  getLine
+
+-- Summary
+type Summary = String
+
+readSummary :: IO Summary
+readSummary = do
+  putStrLn "Give a short summary of the error message."
```
david-christiansen, line 73: I'd also add "This appears on the overview page that lists all the documented errors and warnings"

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+  putStrLn " 1) Error"
+  putStrLn " 2) Warning"
+  putStr "Input (Default = Error): "
+  ln <- getLine
+  case normalize ln of
+    "1" -> pure Error
+    "error" -> pure Error
+    "2" -> pure Warning
+    "warning" -> pure Warning
+    _ -> pure Error
```
david-christiansen, line 92:
```suggestion
   "" -> pure Error
    _ -> do
      putStrLn "Please type \"error\" or \"warning\" or a number."
      readSeverity
```

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+
+-- Warning flag
+type WarningFlag = String
+
+-- | Only ask for a warning flag if Severity = Warning.
+readWarningFlag :: Severity -> IO (Maybe WarningFlag)
+readWarningFlag Warning = do
+  putStrLn "What is the warning flag which enables this warning."
```
david-christiansen, line 100:
```suggestion
  putStrLn "What is the warning flag which enables this warning?"
```

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+  putStrLn "For example, enter \"-Wtabs\" if you are documenting GHC's warning about tabs in your source file."
+  putStr "Input: "
+  Just <$> getLine
```
david-christiansen, line 103: This is also something that many users may trip over. What about adding "Please leave this blank if you're not sure."?

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+
+-- | Only ask for examples if the system is GHC.
+readExamples :: System -> IO Examples
+readExamples GHC = do
+  putStrLn "How many examples should be generated?"
+  putStr "Input: "
+  ln <- getLine
+  case readMaybe ln :: Maybe Int of
+    Nothing -> pure []
+    (Just n) -> forM [1 .. n] readExample
```
david-christiansen, line 131:
```suggestion
    Just n -> forM [1 .. n] readExample
```

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+  print tmpl
+  putStrLn ""
+
+  -- Create the new directory "messages/XXX-NNNNNN/" and "messages/XXX-NNNNNN/index.md"
+  let message_dir = "messages" </> case system tmpl of { GHC -> "GHC-"; GHCup -> "GHCup-"; Stack -> "S-" } ++ code tmpl
+  createDirectory message_dir
+  let toplvl_index =
+        unlines
+          [ "---",
+            "title: " <> title tmpl,
```
david-christiansen, line 192: Do we have to worry about any of the complicated YAML coercion rules here?

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+            "---",
+            "",
+            "Insert your error message here."
+          ]
+  writeFile (message_dir </> "index.md") toplvl_index
+
+  -- Create the example directories and entries:
+  -- - "messages/XXX-NNNNNN/" and "messages/XXX-NNNNNN/index.md"
+  -- - "messages/XXX-NNNNNN/before/" and "messages/XXX-NNNNNN/before/Module.hs"
+  -- - "messages/XXX-NNNNNN/after/" and "messages/XXX-NNNNNN/after/Module.hs"
```
david-christiansen, line 205: This comment seems incorrect to me. Shouldn't there be a `messages/XXX-NNNNN/informative-example-name/index.md` and so on?

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+readVersion = do
+  putStrLn "Which version of the tool emitted the numeric code (not the message) for the first time?"
+  putStrLn "Note: For GHC this is most likely 9.6.1."
+  putStr "Input: "
+  getLine
+
+-- Examples
+type Examples = [String]
+
+validateExampleName :: String -> Bool
```
david-christiansen, line 119: I think that this should also encourage example names to be lowercase

message-index/helper-tool.hs
david-christiansen: I think the code is right, just not the comment

message-index/helper-tool.hs
```diff
@@ -0,0 +1,240 @@
...
+              "---"
+            ]
+    writeFile (example_dir </> "index.md") example_index
+    let before_module =
+          unlines
+            [ "module Example where",
+              "",
+              "-- Insert the example containing a bug here."
+            ]
+    writeFile (example_dir </> "before" </> "Example.hs") before_module
```
david-christiansen, line 224: What about basing the file name `Example.hs` on the `example` string? Something like uppercasing the first letter and adding `.hs`? This makes it easier for people who download the code and work with it to keep track

david-christiansen (changes requested):
A few fairly small comments.

I think that this is also much more useful if it includes easy instructions for running it.

david-christiansen:
Thanks for this, I think it really lowers the barriers to contribution!

message-index/helper-tool.hs
BinderDavid: Done in [aaa4dec](https://github.com/haskellfoundation/error-message-index/pull/408/commits/aaa4deca07075d4eb49e9acb9434ff564ddfa289)

message-index/helper-tool.hs
BinderDavid: Done in [aaa4dec](https://github.com/haskellfoundation/error-message-index/pull/408/commits/aaa4deca07075d4eb49e9acb9434ff564ddfa289)

message-index/helper-tool.hs
BinderDavid: Done in [aaa4dec](https://github.com/haskellfoundation/error-message-index/pull/408/commits/aaa4deca07075d4eb49e9acb9434ff564ddfa289)

message-index/helper-tool.hs
BinderDavid: I am not sure about what YAML accepts or doesn't accept, and how to test this in the script without writing complex validation logic :man_shrugging: I hope that incorrect YAML will be caught by the CI when trying to build the site.

BinderDavid:
I applied all of the changes you suggested in the commit [aaa4dec](https://github.com/haskellfoundation/error-message-index/pull/408/commits/aaa4deca07075d4eb49e9acb9434ff564ddfa289) except for the YAML validation, which would probably be quite complex. 

W.r.t documenting the usage of this tool, I hope that we can handle this in #409

message-index/helper-tool.hs
david-christiansen: I think that incorrect YAML is more likely to create icky output HTML than it is to actually cause an error message, unfortunately. But we can always add quotes around things if we need to afterwards.

david-christiansen (approved):

david-christiansen:
I would prefer a more descriptive name like @ulysses4ever, but I don't think it's a blocker. Better to get this in here prior to Zurihac!

BinderDavid:
> I would prefer a more descriptive name like @ulysses4ever, but I don't think it's a blocker. Better to get this in here prior to Zurihac!

Renamed it to `create-message-template.hs` :)

david-christiansen:
:fireworks:

PR #410, opened 2023-04-21
Fix permission and paths for HLint action.

chungyc:
* GitHub requires the write permission for `security-events` to upload analysis results.  See https://docs.github.com/en/rest/code-scanning#upload-an-analysis-as-sarif-data

* Paths are passed to HLint as is, which does not use them as file patterns.  The paths were changed to a specific path instead of a file pattern for consistency.  In addition, there was a typo in a path.

To see the HLint action in action in a pull request, see https://github.com/chungyc/error-message-index/pull/2.

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -18,6 +18,9 @@ jobs:
   hlint:
     name: HLint
     runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload results to GitHub code scanning.
+      security-events: write
     steps:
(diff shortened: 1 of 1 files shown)


david-christiansen:
Thanks! All this sounds good, but I know very little about GitHub Actions. @BinderDavid, can you take a peek? Thanks!

BinderDavid (approved):
Yes, this looks good. I discovered that the hlint action doesn't work correctly in #408, but didn't have any time to look at the problem in the meantime. Thanks for fixing this.

PR #414, opened 2023-05-30
Test for HLint code scanning.

chungyc:
This is to confirm and demo HLint code scanning working with a pull request against a repository for which the pull request author has no write permissions for.

PR #415, opened 2023-05-30
Scan HLint on pull requests.

chungyc:
According to https://github.com/orgs/community/discussions/54013, uploading scan results for a pull request no longer requires write permission against the destination repository. Re-enable HLint code scanning for pull requests, which will also show any warnings and corresponding code in the pull request itself.

A sample pull request is at https://github.com/haskellfoundation/error-message-index/pull/414.


BinderDavid (approved):
LGTM  Thanks :)

PR #419, opened 2023-06-10
Print warning flag in generated index.md

BinderDavid:
The script asks for a warning-flag, but doesn't print it when we are generating a `index.md` file.


david-christiansen (approved):
LGTM!

PR #420, opened 2023-06-10
Allow pandoc-types 1.23

isomorpheme:
Apparently Cabal in CI could do the version bump from #417, but not in my local setup? So this update seems to fix it, it's probably downstream from Hakyll using a newer Pandoc.

### DIFF
--- a/message-index/message-index.cabal
+++ b/message-index/message-index.cabal
@@ -12,7 +12,7 @@ executable site
                   , microlens ^>= 0.4.12
                   , binary ^>= 0.8.8
                   , aeson ^>= 2.0.3 || ^>= 2.1
-                  , pandoc-types ^>= 1.22
+                  , pandoc-types ^>= 1.22 || ^>= 1.23
                   , containers ^>= 0.6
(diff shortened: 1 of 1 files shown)


BinderDavid (approved):
LGTM I observed the same problem on my machine, so I can reproduce :)

message-index/message-index.cabal
```diff
@@ -12,7 +12,7 @@ executable site
                   , microlens ^>= 0.4.12
                   , binary ^>= 0.8.8
                   , aeson ^>= 2.0.3 || ^>= 2.1
-                  , pandoc-types ^>= 1.22
+                  , pandoc-types ^>= 1.23
```
david-christiansen, line 15: What if this were instead `pandoc-types ^>= 1.22 ||  ^>= 1.23`?

isomorpheme: Good idea, I've done that & changed the commitmsg/PR title accordingly.

david-christiansen (approved):

PR #421, opened 2023-06-10
Explain GHC-39999 arising from OverloadedRecordDot

simmsb:
This adds an explanation for the instance resolution failure that arises when using overloaded record dot and accessing a field of a known type when the field is not in scope


david-christiansen (approved):
This is great!

PR #423, opened 2023-06-10
Add docs for GHC-04584

### LINKED ISSUE #278: PsErrArrowExprInPat -> 4584

TauOmicronMu:
Closes #278

Think this reads alright :)


message-index/messages/GHC-04584/matchingdots/index.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Using .. in pattern 
+---
+
+When pattern matching, Expressions are not allowed as patterns to be matched against. Instead, consider matching on a list, and using a conditional.
```
BinderDavid, line 5: Expressions > expressions

message-index/messages/GHC-04584/index.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Expression syntax in pattern
+summary: There has been an attempt to pattern match on Expression syntax.
```
BinderDavid, line 3: Expression > expression

BinderDavid (changes requested):
Looks very good :) I think there are just 2 remaining small changes in capitalization.

message-index/messages/GHC-04584/index.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Expression syntax in pattern
+summary: There has been an attempt to pattern match on Expression syntax. 
+severity: error
+introduced: 9.6.1
+---
+
+## Example Text
```
david-christiansen, line 8: Normally we have a general description of the error here, and then specific instances of the error included with the programs that generate them.

I'd use a text like "Pattern matching allows constructors and literals to be used to take apart values, exposing their contents. The syntax of patterns is very similar to that of the expressions that create values, but many constructs that are allowed in expression contexts are not meaningful patterns. "

david-christiansen (changes requested):
There's a general style thing to make it consistent with the rest of the docs. The example is really well-chosen.

message-index/messages/GHC-09646/index.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Tuple section in pattern
+summary: There has been an attempt to pattern match on a tuple section.
+severity: error
+introduced: 9.6.1
+---
+
+Pattern matching allows constructors and literals to be used to take apart values, exposing their contents. The syntax of patterns is very similar to that of the expressions that create values, but many constructs that are allowed in expression contexts are not meaningful patterns.
```
david-christiansen: Here, I'd use a more specific text for this specific variation.

```suggestion
[Tuple sections](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/tuple_sections.html) are a concise syntax for functions that result in tuples. The missing expressions in the tuple become arguments to the function. Because tuple sections always result in functions, and functions cannot be matched in patterns, tuple sections are not allowed in patterns.
```

message-index/messages/GHC-09646/snd/index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Using tuple section in pattern
+---
+
+When pattern matching, tuple sections are not allowed as patterns to be matched against. In the example below, we can fix this by adding the wildcard _ to match on the first element of the pair.
```
david-christiansen: I would explain the intent behind the example more. E.g.
```suggestion
When pattern matching, tuple sections are not allowed as patterns to be matched against. In the example below, one of the components of a tuple pattern was forgotten. We can fix this by adding the wildcard _ to match on the first element of the pair.  
```

david-christiansen (changes requested):

david-christiansen:
Looks like we had a review race condition!

message-index/messages/GHC-09646/index.md
TauOmicronMu: Have removed the GHC-09646 stuff from this PR and put it onto a new branch - didn't mean for it to come through here. I'll make the requested changes before sending the new PR over though.

david-christiansen (approved):

message-index/messages/GHC-09646/index.md
david-christiansen: Thanks!

david-christiansen:
@BinderDavid I'll let you approve before a merge

BinderDavid (approved):

PR #427, opened 2023-06-11
document GHC-55666

HugoPeters1024:
The warning is not emitted by default, hence I first thought to have found a bug in GHC. Luckily that was not the case but I my inexperienced fingers still yearned for a contribution so I thought I'd add the warning to the error index instead :)

### DIFF
--- a/message-index/messages/GHC-55666/example/after/Bang_on_unlifted_type.hs
+++ b/message-index/messages/GHC-55666/example/after/Bang_on_unlifted_type.hs
@@ -1,0 +1,5 @@
+module Bang_on_unlifted_type where
+
+import GHC.Base (Int#)
+
+data T = MkT Int#
--- a/message-index/messages/GHC-55666/example/index.md
+++ b/message-index/messages/GHC-55666/example/index.md
@@ -1,0 +1,3 @@
+---
+title: Bang (!) on unlifted type
+---
--- a/message-index/messages/GHC-55666/index.md
+++ b/message-index/messages/GHC-55666/index.md
@@ -1,0 +1,16 @@
+---
+title: Strictness annotation on unlifted type
+summary: Using a strictness annotation (bang) on an unlifted type is redudant as unlifted values are strict by definition
+severity: warning
+flag: -Wredundant-strictness-flags
+introduced: GHC 9.6.1
+---
+
+A strictness annotation (also called a bang: `!`) can be used to denote that a type should not be evaluated lazily.
+In some cases this can lead to faster code because fewer heap allocation are required.
+
+However, unlifted types like `Int#` are strict by definition because they are a value,
+not a pointer to a potentially unevaluated value (thunk).
+
+Therefore, adding strictness annotations to unlifted types or fields of such types are redudant.
+They should be omitted to prevent confusion.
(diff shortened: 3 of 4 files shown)


message-index/messages/GHC-55666/index.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Strictness annotation on unlifted type
+summary: Using a strictness annotation (bang) on an unlifted type is redudant as unlifted values are strict by definition
+severity: warning
+flag: -Wredundant-strictness-flags
+introduced: GHC 9.6.1
+---
+
+A strictness annotation (also called a bang: `!`) can be used to denote that a type should not be evaluated lazily.
```
jhrcek, line 9: This could lead to some confusion, because there's not such thing as "evaluating a type". Types are actually erased during compilation and only description of value-level computation (my amateurish formulation) remain.

I suggest reformulating this to something like ".. can be used to mark fields within data types, that should not be evaluated lazily".

message-index/messages/GHC-55666/index.md
```diff
@@ -0,0 +1,16 @@
...
+In some cases this can lead to faster code because fewer heap allocation are required.
+
+However, unlifted types like `Int#` are strict by definition because they are a value,
+not a pointer to a potentially unevaluated value (thunk).
+
+Therefore, adding strictness annotations to unlifted types or fields of such types are redudant.
```
jhrcek, line 15:
```suggestion
Therefore, adding strictness annotations to unlifted types or fields of such types is redundant.
```

message-index/messages/GHC-55666/index.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Strictness annotation on unlifted type
+summary: Using a strictness annotation (bang) on an unlifted type is redudant as unlifted values are strict by definition
+severity: warning
+flag: -Wredundant-strictness-flags
+introduced: GHC 9.6.1
+---
+
+A strictness annotation (also called a bang: `!`) can be used to denote that a value should not be evaluated lazily.
+In some cases this can lead to faster code because fewer heap allocation are required.
```
david-christiansen, line 10:
```suggestion
In some cases this can lead to faster code because fewer heap allocations are required.
```

message-index/messages/GHC-55666/example/after/Bang_on_unlifted_type.hs
```diff
@@ -0,0 +1,5 @@
+module Bang_on_unlifted_type where
```
BinderDavid, line 1: We usually add the Language extensions which are needed to parse the file as a Language Pragma here:
So I would add `{-# LANGUAGE MagicHash #-} both in the before and the after example.

message-index/messages/GHC-55666/example/index.md
```diff
@@ -0,0 +1,3 @@
+---
+title: Bang (!) on unlifted type
+---
```
BinderDavid, line 3: Could you add the error message from your example inside code fences in this file?

message-index/messages/GHC-55666/index.md
```diff
@@ -0,0 +1,17 @@
...
+Here it is used to mark a that a field in a datatype should not be evaluated lazily.
+
+However, unlifted types like `Int#` are strict by definition because they are a value,
+not a pointer to a potentially unevaluated value (thunk).
+
+Therefore, adding strictness annotations to unlifted types or fields of such types are redundant.
```
david-christiansen, line 16:
```suggestion
Therefore, adding strictness annotations to unlifted types or fields of such types is redundant.
```

david-christiansen (changes requested):
Two minor language tweaks - but this looks good!

david-christiansen (approved):

PR #434, opened 2023-06-11
Document GHC-00711

BinderDavid:
Companion PR: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10636/diffs#c2554d4efde41e6e193f728122336449d2cb5fc5 :)

### DIFF
--- a/message-index/messages/GHC-00711/index.md
+++ b/message-index/messages/GHC-00711/index.md
@@ -1,0 +1,9 @@
+---
+title: Duplicate warning declarations
+summary: Duplicate deprecation warnings are illegal
+severity: error
+introduced: 9.6.1
+---
+
+Using a deprecation pragma it is possible to declare that a Haskell entity such as a type or a toplevel function is deprecated.
+It is, however, illegal to specify multiple deprecation warnings for the same entity.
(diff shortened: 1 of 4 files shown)


message-index/messages/GHC-00711/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Duplicate warning declarations
+summary: Duplicate deprecation warnings are illegal
+severity: error
+introduced: 9.6.1
+---
+
+Using a deprecation pragma it is possible to declare that a Haskell entity such as a type or a toplevel function is deprecated.
```
david-christiansen, line 8:
```suggestion
A Haskell entity, such as a type or a top-level definition, can be declared to be deprecated using a pragma.
```
or
```suggestion
Using a deprecation pragma, it is possible to declare that a Haskell entity such as a type or a top-level function is deprecated.
```

I prefer the former fix, as I feel that it reads a bit less repetitively. But either fix is just fine - pick what you prefer!

david-christiansen (changes requested):
I have a grammar/spelling fix, or an alternative phrasing for the text in question. Otherwise, LGTM. Thanks!

BinderDavid:
I adapted your suggested change, thanks! I will merge  now :)

PR #436, opened 2023-06-12
Documenting two examples for `83865`; Adds `order` field to example headers

### LINKED ISSUE #438: Optional explicit ordering of examples
Talking with @VictorCMiraldo at lunch today about #436, we came to the conclusion that error docs should support optional ordering of examples (when the one writing the documentation has a particular idea about which examples are most likely to be relevant). The design we hit on would be extending the YAML block in an example's `index.md` to allow an optional field `order` that should contain an int. Examples can then be ordered by this field. Those without an order would come at the end (so absence is negative infinity) and anything with the same explicit `order` field would be displayed in some arbitrary order within the group.

VictorCMiraldo:
That's a tricky error code, it shows up in many situations (https://gitlab.haskell.org/ghc/ghc/-/issues/23466), but at the same time it is one of the errors that beginners will hit the most. I attempted to document what I believe are the two situations where 83865 will appear first when someone is learning Haskell, so I tried to keep the examples almost trivially simple and the explanations short and concise. What do you folks think?

Closes #438

### DIFF
--- a/message-index/messages/GHC-83865/index.md
+++ b/message-index/messages/GHC-83865/index.md
@@ -1,0 +1,15 @@
+---
+title: Type Mismatch
+summary: You provided a value of a given type, whereas GHC expected a different type.
+severity: error
+introduced: 9.6.1
+---
+
+GHC expected one type, but was given another. Unlike dynamically-typed programming languages,
+type signatures in Haskell are like a contract between the programmer and the compiler.
+In its simplest form, when you call a function `f` with type `a -> b`, with some argument `x`,
+the compiler will check whether `x` has type `a` and if that is not the case, it will trigger
+the type mismatch error. This case is illustrated by the `terms` example, below.
+
+Type mismatch errors are quite general, however, so you will still encounter them
+in many other situations.
--- a/message-index/messages/GHC-83865/terms/index.md
+++ b/message-index/messages/GHC-83865/terms/index.md
@@ -1,0 +1,15 @@
+---
+title: Values of Different Types
+---
+
+Function `inc` has type `Int -> Int`, hence it expects an argument
+of type `Int`; yet, on the definition of `two` it was called with
+an argument of type `String`.
+
+If you ever need to know the type of something, you can ask for it in `ghci`
+and ask for it with `:type` (or its shorthand `:t`):
+
+```
+ghci> :t "x"
+"x" :: String
+```
--- a/message-index/messages/GHC-83865/terms/index.md
+++ b/message-index/messages/GHC-83865/terms/index.md
@@ -8,7 +8,7 @@ of type `Int`; yet, on the definition of `two` it was called with
 an argument of type `String`.
 
 If you ever need to know the type of something, you can ask for it in `ghci`
-and ask for it with `:type` (or its shorthand `:t`):
+with the command `:type` (or its shorthand `:t`):
 
 ```
 ghci> :t "x"
--- a/message-index/messages/GHC-83865/terms/index.md
+++ b/message-index/messages/GHC-83865/terms/index.md
@@ -1,5 +1,6 @@
 ---
 title: Values of Different Types
+order: 1
 ---
 
 Function `inc` has type `Int -> Int`, hence it expects an argument
--- a/message-index/messages/GHC-83865/terms/index.md
+++ b/message-index/messages/GHC-83865/terms/index.md
@@ -1,6 +1,6 @@
 ---
 title: Values of Different Types
-order: 1
+order: 0
 ---
 
 Function `inc` has type `Int -> Int`, hence it expects an argument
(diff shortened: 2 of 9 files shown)


message-index/messages/GHC-83865/index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Type Mismatch
+summary: You a value of a given type, whereas GHC expected a different type.
```
jhrcek, line 3: Seems like a verb is missing (You <what> a value of a given type)

VictorCMiraldo: Fixed, thanks! :)

VictorCMiraldo:
I'm trying to address #438 as part of this PR, since this particular error is a great test case for it, but it seems like Hakyll insists in ignoring the order of the list it is given. I didn't have time to dig into Hakylls source code to check that yet. @david-christiansen, would you have an idea?

david-christiansen:
I'm stuck in travel mode (flight chaos), so I'm not sure when I can look, but I will do it when I can! We sort lists all the time in the HF website code, so it should be possible.

david-christiansen:
Thanks for looking into this!

VictorCMiraldo:
Thanks for the quick reply, I'll try a little longer tomorrow! Safe travels

VictorCMiraldo:
@david-christiansen turns out I was trying to sort the wrong thing, I got it working this morning! :tada:

message-index/messages/GHC-83865/terms/index.md
```diff
@@ -0,0 +1,31 @@
...
+title: Values of Different Types
+order: 0
+---
+
+Function `inc` has type `Int -> Int`, hence it expects an argument
+of type `Int`; yet, on the definition of `two` it was called with
+an argument of type `String`.
+
+If you ever need to know the type of something, you can ask for it in `ghci`
+and ask for it with `:type` (or its shorthand `:t`):
```
david-christiansen, line 11:
```suggestion
with the command `:type` (or its shorthand `:t`):
```

message-index/messages/GHC-83865/type-kind-mism/index.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Type expected, but kind received.
+order: 1
+---
+
+Forgetting the type parameter to `Maybe` is the culprit, but it is only caught in the
+context of the the arrow in the declaration of `isNothing`, which can be confusing.
+Recall that the arrow (`->`) in Haskell is a type constructor, it takes two types
```
david-christiansen, line 8:
```suggestion
The arrow (`->`) in Haskell is a _type constructor_. It takes two types
```
I'm not a fan of "recall" in this context, because it implies that a reader should already know this information. But I can imagine many situations where exactly this document is someone's first language documentation for Haskell, so we can't expect them to.

message-index/site.hs
```diff
@@ -193,6 +195,18 @@ messageTitleField = field "title" getTitle
 messageCtx :: Context String
 messageCtx = field "id" (pure . getId) <> indexlessUrlField "url"
 
+data ExampleOrder = InGroup Integer | Last
+  deriving (Eq, Ord, Show)
```
david-christiansen, line 199: I think that this code would be a bit more clear if the `Ord` instance were not derived. This is because we're not using `Ord` just to stick the value into a binary tree or something - it's being used for the precise semantics of what is bigger than what.
```suggestion
  deriving (Eq, Show)

instance Ord ExampleOrder where
  compare (InGroup i) (InGroup j) = compare i j
  compare (InGroup _) Last = LT
  compare Last (InGroup _) = GT
  compare Last Last = EQ  
```

david-christiansen (changes requested):
Thanks for this!

I have some minor nitpicks - what do you think?

message-index/messages/GHC-83865/type-kind-mism/index.md
VictorCMiraldo: That makes a lot of sense!

message-index/site.hs
VictorCMiraldo: Very fair! Plus if someone ever changes the order of constructors by accident, they don't mess up the orderings on the website.

VictorCMiraldo:
Thank you for the suggestions and for writing them as clickable suggestions, @david-christiansen! I agree with all and have had them applied!

message-index/site.hs
```diff
@@ -193,6 +195,24 @@ messageTitleField = field "title" getTitle
...
 messageCtx = field "id" (pure . getId) <> indexlessUrlField "url"
 
+data ExampleOrder = InGroup Integer | Last
+  deriving (Eq, Show)
+
+instance Ord ExampleOrder where
+  compare (InGroup i) (InGroup j) = compare i j
+  compare (InGroup _) Last = LT
+  compare Last (InGroup _) = GT
+  compare Last Last = EQ
```
VictorCMiraldo, line 205:
```suggestion
  compare Last Last = EQ
```

BinderDavid (approved):
The ordering seems to work on my computer, and you adressed all the requested changes w.r.t to the wording. So this now looks good to me 👍

PR #437, opened 2023-06-12
Replace highlight.js with static highlighting using Pandoc 

TauOmicronMu:
🥳 🎉


message-index/site.hs
```diff
@@ -76,13 +77,9 @@ main = hakyll $ do
...
                       [ indexlessUrlField "url",
                         field "name" (pure . view _1 . itemBody),
-                        -- Set the language that highlight.js should use for syntax highlighting
-                        field "language" $ \(itemBody -> (filename, _, _)) ->
-                          pure $ case dropWhile (== '.') $ takeExtension filename of
-                            "hs" -> "haskell"
-                            other -> other,
-                        field "before" (maybe (pure "<not present>") (fmap itemBody . load . itemIdentifier) . view _2 . itemBody),
-                        field "after" (maybe (pure "<not present>") (fmap itemBody . load . itemIdentifier) . view _3 . itemBody)
+                        -- TODO: pick the right language
```
david-christiansen, line 80: Is this a stale TODO?

message-index/site.hs
```diff
@@ -76,13 +77,9 @@
...
+                        field "before" (maybe (pure "<not present>") (fmap (T.unpack . highlightHaskell . T.pack) . fmap itemBody . load . itemIdentifier) . view _2 . itemBody),
+                        field "after" (maybe (pure "<not present>") (fmap (T.unpack . highlightHaskell . T.pack) . fmap itemBody . load . itemIdentifier) . view _3 . itemBody)
```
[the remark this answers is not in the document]
david-christiansen: I wonder if we should just turn this off in an hlint config - it seems to me to make the code harder to understand, not easier

david-christiansen (changes requested):
For Haskell, this looks great to me.

But for other languages, it looks like Haskell highlighting is being used. Here's some YAML:
image

Shouldn't it branch on the code block language, or alternatively on the filename?

thanks!

message-index/site.hs
TauOmicronMu: I'll have a look at it now!

david-christiansen:
Is this still a draft? What do you see as the remaining tasks for un-drafting it?

TauOmicronMu:
@isomorpheme - completely forgot about this -> do you remember what we had left to do (if anything)?

isomorpheme:
@TauOmicronMu Hmm, I don't remember anything besides what's already mentioned in this PR (mainly picking the right language to highlight with). And this needs a rebase to fix merge conflicts of course.

PR #439, opened 2023-06-12
Add instruction to fork first, pending bug fix from GitHub

david-christiansen:
GitHub presently doesn't work when clicking the "Fork" link with default text for a new file. I reported a bug; in the meantime, this PR adds an instruction to fork it first.


david-christiansen:
GitHub let me know that the bug was fixed - this is no longer needed.

PR #441, opened 2023-06-25
Added another example of error GHC-39999

blamario:
I think this example is worth including because it's an effect of a backward incompatible change in `base`.


BinderDavid (approved):
I added an ordering to the examples, a feature which was just merged in #436.
Apart from that, LGTM! Thanks for the example. I will merge as soon as CI runs through again.

PR #445, opened 2023-07-09
Document GHC-97044

jhrcek:


message-index/messages/GHC-97044/index.md
```diff
@@ -0,0 +1,29 @@
...
+
+There are several special type classes in [base](https://hackage.haskell.org/package/base) which require special treatment from the compiler. For this reason they don't permit user-specified instances.
+
+An attempt to provide an explicit instance declaration in your library / application code, e.g.
+```haskell
+instance Coercible A B where ...
+```
+will lead to this error being reported.
+
+The instances for these type classes are automatically created by GHC on as-needed basis.
```
BinderDavid, line 16:
```suggestion
The instances for these type classes are automatically created by GHC on an as-needed basis.
```

message-index/messages/GHC-97044/index.md
```diff
@@ -0,0 +1,29 @@
...
+
+This restriction applies to the following type classes:
+
+- [Coercible](https://hackage.haskell.org/package/base/docs/Data-Coerce.html#t:Coercible) - This class does not have regular instances; instead they are created on-the-fly during type-checking.
+- [HasField](https://hackage.haskell.org/package/base/docs/GHC-Records.html#t:HasField) - This class is used by [OverloadedRecordDot](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/overloaded_record_dot.html) and related extensions.
```
BinderDavid, line 21: At least for the `HasField` typeclass it is possible to create user-written instances. See for example my library here: https://github.com/BinderDavid/tuple-fields/blob/main/src/Data/Tuple/Fields.hs

BinderDavid (changes requested):
Looks good! I have one suggested change in grammar, and I think the `HasField` typeclass does not belong in that list.

message-index/messages/GHC-97044/index.md
jhrcek: Applied.

message-index/messages/GHC-97044/index.md
jhrcek: You're right. Seems like I was misled by browsing around GHC codebase and I interpreted this 

https://github.com/ghc/ghc/blob/2af23f0e84eec0eb30d77134abd99858a02d7a18/compiler/GHC/Tc/Validity.hs#L1460

to mean that HasField can also trigger this.
But after deeper look it seems to be related to other "validations" done on instances of this type class.

As for other typeclasses mentioned, I'm pretty sure they should be there as I saw them mentioned in warnings in ghc test suite: https://github.com/search?q=repo%3Aghc%2Fghc%2097044&type=code

BinderDavid (approved):

PR #448, opened 2023-08-09
Add GHC-25897

tomjaguarpaw:


message-index/messages/GHC-25897/index.md
```diff
@@ -0,0 +1,39 @@
...
+title: GADT pattern match must have a known result type
+summary: A pattern match on a GADT cannot succeed unless GHC knows the result type of the pattern match.
+severity: error
+introduced: 9.6.1
+---
+
+A pattern match on a GADT cannot succeed unless GHC knows the result
+type of the pattern match, for example because the pattern match has a
+type signature, or because the type of the pattern match can be
+inferred from its context.
```
david-christiansen, line 11: Minor nitpick: I think that this "for example" is a bit confusing - my first read was that it was providing examples of reasons that this knowledge was needed, rather than ways in which it could be found. Here's one rephrasing that I think addresses it:

```suggestion
A pattern match on a GADT cannot succeed unless GHC knows the result
type of the pattern match. This information might, for example, be derived from a type
signature, or by type inference due to the context in which the pattern match occurs.
```

message-index/messages/GHC-25897/index.md
```diff
@@ -0,0 +1,39 @@
...
+        \case
+          FA -> True
+          FB -> False
+    • Relevant bindings include
+        foo1 :: F x -> p (bound at test17.hs:13:1)
+    Suggested fix: Consider giving ‘foo1’ a type signature
+   |
+14 |   FA -> True
+   |         ^^^^
+```
```
david-christiansen, line 39: This message is really associated with `example1` rather than with the overall description - what about moving it to `example1/index.md` along with a brief specific description of the  example, along the lines of "This example uses a GADT pattern match in the definition `foo1`, which has no type signature. Adding a type signature causes the example to be accepted."

message-index/messages/GHC-25897/example1/after/Example1.hs
```diff
@@ -0,0 +1,34 @@
...
+  G B = Bool
+
+-- Compiles successfully, because the
+-- result type of the pattern match
+-- is known, from the type signature,
+-- to be 'G x'
+foo3 :: F x -> G x
+foo3 = \case
+  FA -> True
+  FB -> False
```
david-christiansen, line 34: I worry that adding extra content on the "fixed" side will distract from the fix to the error message on the left. Could this be made into a second broken/fixed example at the cost of being a bit repetitive?

david-christiansen (changes requested):
There's a couple small style changes that would be nice - if you don't have the bandwidth, let me know and I can do them.

tomjaguarpaw:
The wording around "for example" is fine by me. Regarding the examples I think it's better if you do them since I think it will take an order of magnitude more time for me to understand your intention than for you to do it. I'm sure I'll be happy with whatever you come up with :)

david-christiansen:
All right - I went ahead and did it. Thanks for the contribution!

PR #451, opened 2023-08-14
Document GHC-13218

### LINKED ISSUE #252: TcRnLinearFuncInKind -> 13218

BinderDavid:
Fixes #252

### DIFF
--- a/message-index/messages/GHC-13218/illegalKindSignature/index.md
+++ b/message-index/messages/GHC-13218/illegalKindSignature/index.md
@@ -3,7 +3,8 @@ title: <insert a title for this example here>
 ---
 
 In this example, the data type declaration for the `Void1` data type was annotated with the kind signature `* %1 -> *`.
-This is not allowed, the kind signature of that data type should be `* -> *`.
+This would be a linear kind, which Haskell does not support - only types may be linear.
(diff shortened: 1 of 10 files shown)


message-index/messages/GHC-13218/illegalKindSignature/index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: <insert a title for this example here>
+---
+
+In this example, the data type declaration for the `Void1` data type was annotated with the kind signature `* %1 -> *`.
+This is not allowed, the kind signature of that data type should be `* -> *`.
```
david-christiansen, line 6:
```suggestion
This would be a linear kind, which Haskell does not support - only types may be linear.
The kind signature of the data type should be `* -> *`.
```

david-christiansen: I think this makes it a bit more clear what's going on.

david-christiansen (changes requested):
Looks good, modulo one minor comment.

Also, these examples are built in a specific order. What about using the field added in https://github.com/haskellfoundation/error-message-index/pull/436 by @VictorCMiraldo to make this explicit, rather than relying on the collation from the filenames?

PR #454, opened 2023-08-23
Document GHC-91028

### LINKED ISSUE #24: CannotUnifyWithPolytype -> 91028

### LINKED ISSUE #453: Include mention of QuickLook impredicativity in CannotUnifyWithPolytype error (91028)
As mentioned in [GHC issue #20169](https://gitlab.haskell.org/ghc/ghc/-/issues/20169#note_521958), the error message with error code `91028` would benefit from a reference to quick look impredicativity. We decided not to suggest `-XImpredicativeTypes` in the error message as it often does not fix the problem, but in the context of the error message index it seems to make more sense to point users towards that extension and the quick look impredicativity paper.

david-christiansen:
Fixes #24
Fixes #453

Feedback on the somewhat long explanatory text is very welcome!

### DIFF
--- a/message-index/messages/GHC-91028/predkinds/after/PredKinds.hs
+++ b/message-index/messages/GHC-91028/predkinds/after/PredKinds.hs
@@ -4,4 +4,4 @@ module PredKinds where
 import Data.Proxy
 
 t :: forall a (z :: forall k. a). Proxy z
-t = undefined
+t = Proxy
--- a/message-index/messages/GHC-91028/predkinds/after/PredKinds.hs
+++ b/message-index/messages/GHC-91028/predkinds/after/PredKinds.hs
@@ -1,0 +1,7 @@
+{-# LANGUAGE ImpredicativeTypes #-}
+module PredKinds where
+
+import Data.Proxy
+
+t :: forall a (z :: forall k. a). Proxy z
+t = undefined
--- a/message-index/messages/GHC-91028/predkinds/before/PredKinds.hs
+++ b/message-index/messages/GHC-91028/predkinds/before/PredKinds.hs
@@ -4,4 +4,4 @@ module PredKinds where
 import Data.Proxy
 
 t :: Proxy (z :: forall k. a)
-t = undefined
+t = Proxy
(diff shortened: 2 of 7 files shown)


david-christiansen:
@sheaf - how does this fit with what you were looking for?

message-index/messages/GHC-91028/predkinds/after/PredKinds.hs
```diff
@@ -0,0 +1,7 @@
+{-# LANGUAGE ImpredicativeTypes #-}
+module PredKinds where
+
+import Data.Proxy
+
+t :: forall a (z :: forall k. a). Proxy z
+t = undefined
```
jhrcek, line 7: I don't think undefined in necessary in this example.
Using Proxy in this case make the example look less confusing in my opinion.

```suggestion
t = Proxy
```

message-index/messages/GHC-91028/predkinds/before/PredKinds.hs
```diff
@@ -0,0 +1,7 @@
+{-# LANGUAGE ImpredicativeTypes #-}
+module PredKinds where
+
+import Data.Proxy
+
+t :: Proxy (z :: forall k. a)
+t = undefined
```
jhrcek, line 7:
```suggestion
t = Proxy
```

jhrcek (approved):
Looks great. I never actually had to work with impredicativity, but the explanations make sense to me.

message-index/messages/GHC-91028/predkinds/after/PredKinds.hs
david-christiansen: That's a good idea. Thanks!

sheaf:
Fantastic explanations, David. Thanks for doing this!

Is there a way to view the rendered HTML, just to check formatting etc? I couldn't find any links to rendered artifacts from the CI run.

BinderDavid:
> Fantastic explanations, David. Thanks for doing this!
> 
> Is there a way to view the rendered HTML, just to check formatting etc? I couldn't find any links to rendered artifacts from the CI run.

I could try releasing the build artefacts, but we should probably make sure that the artifacts get invalidated after some time, lest we use too much CI disk space in the long run.

BinderDavid (approved):
The explanation is great :+1:

PR #455, opened 2023-08-24
Fix issue #444 (hakyll runs out of file handles)

### LINKED ISSUE #444: Build problems on Mac due to max number of file handles
The error message index has grown in size, which is great, but the number of files can now lead to problems when running the site, since we can exhaust the number of open file handles that a process (in our case: Hakyll) is allowed to have. I just ran into this problem on my Mac when running the site locally:

```
 updated messages/GHC-39999/superclass/after/Main.hs
site: _site/messages/GHC-11913/lower-case/after/LowerCase.hs: copyFileWithMetadata:copyFileContents:openFdAt: resource exhausted (Too many open files)
```

By default, the max. number of file handles on my Mac is 256

```
david@Davids-MacBook-Air message-index % launchctl limit maxfiles
	maxfiles    256            unlimited 
```

I'm investigating what the best fix is. Maybe we have to update the instructions for platforms where the file handle limits are too low.

BinderDavid:
Fixes #444
On systems which don't provide enough file handles to processes, building the site with hakyll can fail since hakyll holds on to too many file handles at the same time.
This commit bumps the minimum requirement on hakyll to version 4.16.1.0, which fixes the problem.

I am just running this through CI for the moment, to see if it also works there.

### DIFF
--- a/message-index/message-index.cabal
+++ b/message-index/message-index.cabal
@@ -6,7 +6,7 @@ cabal-version:      2.0
 executable site
   main-is:          site.hs
   build-depends:    base == 4.*
-                  , hakyll ^>= 4.16.0.0
+                  , hakyll ^>= 4.16.1.0
                   , commonmark >= 0.2.2
                   , filepath ^>= 1.4
                   , microlens ^>= 0.4.12


BinderDavid:
Since this hakyll version is just 1 day old on Hackage, people will have to run `cabal update` in order to be able to build. The error message one gets otherwise is known to not be ideal in that case: https://github.com/haskell/cabal/issues/8922

BinderDavid:
@david-christiansen 
You wrote in the issue:
> I suppose we should experiment a bit with bounds to make sure this release works with everything that we want to support, then close this up.

Did you have anything specific in mind which maybe would not work? I also made sure that the site can be built using `stack build`again, so I think we could just bump the lower bound for hakyll in the cabal file. The only issue I can foresee is that  people have to update their cabal index with `cabal update` before they can build the site again, cabal is unfortunately not very helpful in that case.

david-christiansen:
The only thing I had in mind would be if the new Hakyll required a newer `base` or some other library, but a quick skim of the dependencies doesn't reveal any obvious issues.

david-christiansen (approved):

PR #457, opened 2023-08-24
Add CI for Stack

### LINKED ISSUE #456: CONTRIBUTING.md suggests building with stack, but we don't check whether that is possible in CI
#455 will probably run through, but if we merge it then people won't be able to build using stack, which is one possibility suggested in the CONTRIBUTING.md.

We should have a CI action which checks whether the site is still buildable with stack.
I suggest using just `stack build --dry-run` to check if stack is able to find a build-plan. In #455 with the bump to hakyll this would fail with the following error message:

```
stack build --dry-run

Error: [S-4804]
       Stack failed to construct a build plan.
       
       While constructing the build plan, Stack encountered the following errors:
       
       In the dependencies for message-index-0.1.0.0:
           hakyll-4.15.1.1 from Stack configuration does not match ^>=4.16.1.0 
       needed since message-index is a build target.
       
       Some different approaches to resolving this:
       
         * Set 'allow-newer: true' in /home/david/.stack/config.yaml to ignore all version constraints and build anyway.
```

BinderDavid:
Fixes #456

### DIFF
--- a/.github/workflows/stack-ci.yml
+++ b/.github/workflows/stack-ci.yml
@@ -1,0 +1,29 @@
+name: Stack CI
+
+on:
+  push:
+    paths:
+      - 'message-index/site.hs'
+      - 'message-index/message-index.cabal'
+      - 'message-index/stack.yaml'
+      - 'message-index/stack.yaml.lock'
+    branches:
+      - main
+  pull_request:
+    paths:
+      - 'message-index/site.hs'
+      - 'message-index/message-index.cabal'
+      - 'message-index/stack.yaml'
+      - 'message-index/stack.yaml.lock'
+    branches:
+      - main
+
+jobs:
+  stack-ci:
+    name: Stack CI
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


BinderDavid:
I will merge this PR and then run the action by hand from the "actions" tab. I suspect that it will fail with the following message:

```
Error: [S-4804]
       Stack failed to construct a build plan.
       
       While constructing the build plan, Stack encountered the following errors:
       
       In the dependencies for message-index-0.1.0.0:
           hakyll-4.15.1.1 from Stack configuration does not match ^>=4.16.0.0 (latest matching version is 4.16.1.0)
       needed since message-index is a build target.
       
       Some different approaches to resolving this:
       
         * Set 'allow-newer: true' in /home/david/.stack/config.yaml to ignore all version constraints and build anyway.
       
         * Recommended action: try adding the following to your extra-deps in /home/david/GitRepos/error-message-index/message-index/stack.yaml:
       
       - hakyll-4.16.1.0@sha256:2f5984ec2f0bfc066279f61747839c149f90fbc0d4655c3cf51acda2cf7111b6,9845
 ```

PR #461, opened 2023-09-01
Fix #459 by fixing output of overlapping matches in site.hs

### LINKED ISSUE #459: Error messages for GHCup messages are missing header, title and footer
Screenshot 2023-09-01 at 12-32-44 https __errors haskell org

The error messages for GHCup currently look like above. I.e. all the CSS styling is missing.

BinderDavid:


message-index/site.hs
```diff
@@ -95,7 +94,6 @@ main = hakyll $ do
 
   match "messages/*/index.md" $
     version "nav" $ do
-      route $ setExtension "html"
```
BinderDavid, line 98: If I remove this line, then the navigation via the breadcrumbs stops working.

BinderDavid: Wrong line :(

PR #467, opened 2023-09-15
Add example to GHC-88464

MorrowM:
Adds an example to GHC-88464 (Variable not in scope), since that error can also occur when the programmer forgot an import declaration.


BinderDavid (approved):
Thanks. Looks good!

PR #472, opened 2023-11-03
Create index.md

tobz619:
Outlining problem and error code received.

### DIFF
--- a/message-index/messages/GHCup-08841/index.md
+++ b/message-index/messages/GHCup-08841/index.md
@@ -1,0 +1,35 @@
+---
+title: BuildFailure With Compile
+summary: When using ghcup compile hls -g master --ghc 8.10.4; a build failure occurs.
+introduced: ghcup, version 0.1.19.4
+severity: error
+---
+
+```
+$ ghcup compile hls -g master --ghc 8.10.4
+[ Warn  ] New ghc version available. If you want to install this latest version, run 'ghcup install ghc 9.8.1'
+[ Warn  ] New stack version available. If you want to install this latest version, run 'ghcup install stack 2.13.1'
+[ Info  ] Fetching git repo https://github.com/haskell/haskell-language-server.git at ref master (this may take a while)
+[ Info  ] Examining git ref master
+[ ...   ]   HLS version (from cabal file): 2.4.0.0
+[ ...   ]   commit hash: 0f9be3ea7c6b189cfbf8660bf2766eb9180a78ce
+[ Info  ] Building HLS 2.4.0.0 for GHC version 8.10.4
+[ cabal ] ghc-8.4.4, ghc-8.4.3, ghc-8.4.1, ghc-8.2.2, ghc-8.2.1, ghc-9.2.3.20220620
+[ cabal ] (constraint from non-upgradeable package requires installed instance)
+[ cabal ] [__1] fail (backjumping, conflict set: ghc, ghcide)
+[ cabal ] After searching the rest of the dependency tree exhaustively, these were the
+[ cabal ] goals I've had most trouble fulfilling: ghcide, ghc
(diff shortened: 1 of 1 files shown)


message-index/messages/GHCup-08841/index.md
```diff
@@ -0,0 +1,35 @@
+---
+title: BuildFailure With Compile
+summary: When using ghcup compile hls -g master --ghc 8.10.4; a build failure occurs.
+introduced: ghcup, version 0.1.19.4
```
BinderDavid, line 4: The `introduced` field should be just the number, i.e. `0.1.19.4`.

message-index/messages/GHCup-08841/index.md
```diff
@@ -0,0 +1,35 @@
+---
+title: BuildFailure With Compile
+summary: When using ghcup compile hls -g master --ghc 8.10.4; a build failure occurs.
```
BinderDavid, line 3: The summary should probably reflect a more general summary of this error code. This error code can probably occur with more than just this given invocation of ghcup.

message-index/messages/GHCup-08841/index.md
```diff
@@ -0,0 +1,35 @@
...
+[ Error ] [GHCup-08841] BuildFailed failed in dir /home/.../.ghcup/tmp/ghcup-eb35d2cd0af3dbd1: Process "cabal" with arguments ["v2-install",
+[ ...   ]                                                                                                                 "-w",
+[ ...   ]                                                                                                                 "ghc-8.10.4",
+[ ...   ]                                                                                                                 "--install-method=copy",
+[ ...   ]                                                                                                                 "--overwrite-policy=always",
+[ ...   ]                                                                                                                 "--disable-profiling",
+[ ...   ]                                                                                                                 "--disable-tests",
+[ ...   ]                                                                                                                 "--installdir=/home/tobzdaman619/.ghcup/tmp/ghcup-eb35d2cd0af3dbd1/out/8.10.4",
```
BinderDavid, line 30: You probably want to pick a generic user name in this error message instead leaving your own one in place ;)

message-index/messages/GHCup-08841/index.md
```diff
@@ -0,0 +1,35 @@
...
+[ ...   ]                                                                                                                 "--project-file=cabal.project",
+[ ...   ]                                                                                                                 "exe:haskell-language-server",
+[ ...   ]                                                                                                                 "exe:haskell-language-server-wrapper"] failed with exit code 1.
+```
+This occurs with all versions of cabal and ghc so far.
```
BinderDavid, line 35: Thanks for contributing to the error message index! I think we still have to dig up a little bit more information on GHCup which explains under what circumstances this error occurs, so that we can present a good explanation of this error to the user of the error message index. At first glance it looks like this error occurs when ghcup tries to build some tool such as HLS or GHC, and the build process fails for some reason. I think we cannot get around the fact that we have to look at the GHCup codebase itself and find out more there. Do you know which constructor is responsible for this error message, and are there comments in the GHCup codebase which explain when this error is thrown?

BinderDavid (changes requested):
As I wrote in the other comment, we still need to figure out more precisely what this error code stands for, so that we can present the user of the error message index with some actionable advice.

tobz619:
> As I wrote in the other comment, we still need to figure out more precisely what this error code stands for, so that we can present the user of the error message index with some actionable advice.

Haha no worries, I'm new to this and wanted to help in any way I can :)

message-index/messages/GHCup-08841/index.md
tobz619: Unfortunately, I have no clue :( I've posted every error I get in the original post. Where do you recommend I start looking for clues?

PR #473, opened 2023-12-01
Add missing GHC messages

### LINKED ISSUE #442: Do undocumented error number pages simply not exist?
A user [posted an error message containing an error code](https://discourse.haskell.org/t/cabal-quickstart-with-hls-pains/6867). It just so happens that it's `GHC-32850` (although the specific code doesn't matter for this issue). Google returns exactly one result for `"GHC-32850"` which is [a GHC GitLab issue containing a similar error message](https://gitlab.haskell.org/ghc/ghc/-/issues/23355).

I was surprised. I expected there to be a page for the error code in question, even if it says "this error message is not yet documented". In fact, that seems like an essential workflow for inexperienced users. Googling the code and getting no useful hits means that they haven't been helped by the index at all. On the other hand, if Google returns a link to a stub page that says "This error code is undocumented: please click here to file a GitHub issue to request documentation" that's really helpful!

So,

1. is it really the case that there are error codes that don't have pages, or have I missed something?
2. if so, what can I do to make sure they all get pages, even if they're just stubs?

xsebek:
* generate "Not documented" stub pages for known GHC error codes

This is done by using `GHC.Types.Error.Codes.constructorCodes` by @sheaf from https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10857 and making an `index.md` for each code using a script.

* closes #442


message-index/messages/GHC-00828/index.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Not documented
+---
+
+Message `GHC-00828` is not yet documented.
+Tip: this message has an internal GHC name `DsRuleIgnoredDueToConstructor`.
+
+You can help out by documenting it. There are two ways to get started:
+* [Submit documentation directly through GitHub's online editor](https://github.com/haskellfoundation/error-message-index/new/main/message-index/messages/GHC-0048/\?filename\=index.md\&value\=---%0Atitle%3A%20SHORT%20TITLE%0Asummary%3A%20ONE%20SENTENCE%20SUMMARY%0Aintroduced%3A%209.6.1%0Aseverity%3A%20CHOOSE%20ONE%3A%20warning%20error%0A---%0A%0APlease%20describe%20the%20meaning%20of%20the%20error%20here%2C%20formatted%20in%20Markdown.)
+* [Add documentation or examples on your own computer](https://github.com/haskellfoundation/error-message-index/blob/main/CONTRIBUTING.md) and send a pull request
```
xsebek, line 10: Seems I overlooked how good the current 404 template is - if this was used instead, then the GHC number inside the online editor link needs to be generated too.

BinderDavid:
Thanks a lot for taking up this issue :) It is important to get this done 👍 

I am not yet convinced that adding the missing errors as markdown `index.md` files to the repository is the technically best solution. This approach will make it more difficult if we want to change the wording for any of the missing errors since we have to write a new script which modifies all these files.
It will also make it more difficult to add new missing errors that will be added in future releases of GHC, since the script that you used to generate these messages is not part of the repo.

But hakyll which we use as a static site generator is super flexible when it comes to generating pages from arbitrary  functions written in Haskell. So I think we can just rebuild the logic of your script within  our `site.hs` and add one single file `MissingErrors.txt` to the repository which contains the output of the ghc script. In that way we can easily extend the list of missing error messages, and we can also programmatically change how we want to present missing error messages.

PR #477, opened 2024-02-23
Document GHC-61689; Add examples to GHC-88464 and GHC-76037

tbidne:
Hello!

I recently ran into the error like

```
E2NoFieldSelectors.hs:11:24: error: [GHC-61689]
    Module ‘A’ does not export ‘fint’.
    Suggested fix:
      Notice that ‘fint’ is a field selector belonging to the type ‘A.Foo’
      that has been suppressed by NoFieldSelectors.
   |
11 | import A (Foo (MkFoo), fint)
   |                        ^^^^
```

And I found the various nuances confusing enough that I thought it merited a description here. It turns out there are multiple errors that can trigger this message.

### GHC-61689 (New)

This is a new error introduced in GHC 9.8.1 for importing a an unexported symbol.

```haskell
-- GHC-61689, fint is a field selector disabled with NoFieldSelectors, always an error
import A (Foo (MkFoo), fint)
```

### GHC-88464

This was previously documented in other cases, but it can also trigger the `NoFieldSelectors` message in the right circumstances.

```haskell
-- GHC-88464, error if fint is used as a selector, OK otherwise
import A (Foo (MkFoo, fint))

getFooInt :: Foo -> Int
getFooInt = fint -- this is the problem
```

### GHC-76037

Similar to GHC-61689, though it involves the _export_, not the import.

```haskell

{-# LANGUAGE NoFieldSelectors #-}

-- GHC-76037, error if foo is exported at the top-level rather than part of T.
module NoFieldSelectorsExport (T, foo) where

data T = MkT { foo :: T }
```

I have this in the same PR as they are all related, but please let me know if they should be split up, or if there should be issues created. Thanks!

### DIFF
--- a/message-index/messages/GHC-61689/example1/after/A.hs
+++ b/message-index/messages/GHC-61689/example1/after/A.hs
@@ -1,0 +1,7 @@
+module A (one) where
+
+one :: ()
+one = ()
+
+two :: ()
+two = ()
--- a/message-index/messages/GHC-61689/example1/after/SymbolNotExported.hs
+++ b/message-index/messages/GHC-61689/example1/after/SymbolNotExported.hs
@@ -1,0 +1,3 @@
+module SymbolNotExported where
+
+import A (one)
--- a/message-index/messages/GHC-61689/example1/before/A.hs
+++ b/message-index/messages/GHC-61689/example1/before/A.hs
@@ -1,0 +1,7 @@
+module A (one) where
+
+one :: ()
+one = ()
+
+two :: ()
+two = ()
--- a/message-index/messages/GHC-61689/example1/before/SymbolNotExported.hs
+++ b/message-index/messages/GHC-61689/example1/before/SymbolNotExported.hs
@@ -1,0 +1,3 @@
+module SymbolNotExported where
+
+import A (one, two)
--- a/message-index/messages/GHC-61689/example1/index.md
+++ b/message-index/messages/GHC-61689/example1/index.md
@@ -1,0 +1,18 @@
+---
+title: Importing an unexported symbol.
+---
+
+## Message
+```
+SymbolNotExported.hs:3:16: error: [GHC-61689]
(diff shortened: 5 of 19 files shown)


message-index/messages/GHC-61689/e1SymbolNotExported/A.hs
```diff
@@ -0,0 +1,7 @@
+module A (one) where
+
+one :: ()
+one = ()
+
+two :: ()
+two = ()
```
jhrcek, line 7: Please add missing newlines at the end of the file (wherever you see this red circle in github ui - 9x in this PR)
```suggestion
two = ()

```

message-index/messages/GHC-61689/e1SymbolNotExported/after/E1SymbolNotExported.hs
```diff
@@ -0,0 +1,12 @@
+module E1SymbolNotExported where
+
+-- A.hs
```
jhrcek, line 3: Please remove this commented out code - since the code of module A.hs will be displayed on the page in full, I don't see a reason to repeat it in the "importing" modules.

jhrcek (commented):
The examples and explanations look good. Few minor suggestions.

message-index/messages/GHC-61689/e1SymbolNotExported/after/E1SymbolNotExported.hs
tbidne: Ah, I put a single `A.hs` in a higher dir so both examples could share it. But it doesn't show up on the site, hence the comments.

Your idea is better; `A.hs` has been moved to both `before` and `after`, so it shows up now.

jhrcek (approved):
Thanks! Looks good to me.

tbidne:
Thanks for the review!

BinderDavid (approved):

PR #482, opened 2024-03-15
Fix breadcrumbs

Lysxia:
Issue: The breadcrumbs should link to the home page, but they currently link to the current page (see for example https://errors.haskell.org/messages/GHC-00158/)

The cause is that the `breadcrumbsField` function looks up the `"nav"` version of the parents, and then `indexlessUrlField` looks up their URL using `getRoute`, but the `"nav"` versions of the pages have no route.

The fix is to use the default version of the parents instead in `breadcrumbsField`.

Question: I removed an unused `breadcrumbsField` in the `messages/examples/` pages. Is there any use for those pages beyond being later included in the corresponding message pages? If not, I think we can also remove their `route` command.

List of changes:

1. Remove now unused `"nav"` versions of the pages. I am guessing that the original reason for adding `"nav"` versions was to avoid a circular dependency (so that the message pages can link to the homepage which links to the messages). The dependency was introduced by the `load` function. But we can get the URL and title for the breadcrumbs without `load`.
2. Remove the `breadcrumbFields` function which becomes redundant with `breadcrumbsCtx`. On top of creating the "parents" field for the breadcrumbs, it added a `messageTitleField` which appends the [GHC-XXXXXX] identifier to the title. This was used in the messages pages and also the home page, where it just leaves the title unchanged. But that title was already available in `defaultContext`. Instead we add `messageTitleField` only in the messages pages and in the breadcrumbs (just in case, this is currently unused).
3. Simplify `breadcrumbCtx` to only get the url and title.
4. Refactor `indexlessUrlField` by reusing `urlField`.
5. Refactor `indexless` with an auxiliary `stripSuffix`.
6. Remove unused `breadcrumbField` in the `messages/examples/` pages.

### DIFF
--- a/message-index/site.hs
+++ b/message-index/site.hs
@@ -12,7 +12,7 @@ import Data.Data (Typeable)
 import Data.Foldable (for_)
 import Data.Function (on)
 import Data.Functor ((<&>))
-import Data.List (find, isPrefixOf, lookup, nub, sort, sortBy)
+import Data.List (find, lookup, nub, sort, sortBy, stripPrefix)
 import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
@@ -54,10 +54,6 @@ main = hakyll $ do
       route idRoute
       compile copyFileCompiler
 
-  match "messages/*/*/index.md" $
-    version "nav" $ do
-      compile getResourceBody
-
   match "messages/*/*/index.md" $ do
     route $ setExtension "html"
     compile $ do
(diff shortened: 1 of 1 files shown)


BinderDavid:
Thanks! I will try it out on my machine and then approve and merge if it works as expected.
It might very well have been me who introduced this bug. In https://github.com/haskellfoundation/error-message-index/pull/461/files I fixed https://github.com/haskellfoundation/error-message-index/issues/459 . But it seems that in doing so I might have introduced this bug.

BinderDavid (approved):
Looks good! I tested this locally and I verified that it fixes the bug with the wrong backlink in the breadcrumbs.

BinderDavid:
> Question: I removed an unused breadcrumbsField in the messages/examples/ pages. Is there any use for those pages beyond being later included in the corresponding message pages? If not, I think we can also remove their route command.

@Lysxia I am not sure. I did not write the original implementation of `site.hs`, so I don't know what the intent was when specific parts were introduced. If it can be removed without changing the observable behaviour of the site, then we can also remove it. But I am going to merge this first, since your PR fixes an observable bug.

PR #487, opened 2024-05-31
Add example from LYAH to GHC-39999

Bodigrim:
I've been re-reading https://learnyouahaskell.github.io/starting-out.html#ready-set-go and noticed that learners can reasonably be confused why the error in the recent GHCs points to `5` not to `"4"`.


message-index/messages/GHC-39999/adding-number-to-string/index.md
```diff
@@ -0,0 +1,25 @@
...
+      In the second argument of ‘($)’, namely ‘5 + "4"’
+      In the expression: print $ 5 + "4"
+  |
+1 | main = print $ 5 + "4"
+  |                ^
+```
+
+## Explanation
+
+Given that Haskell is a strong statically typed language, it is not a surprise that adding numbers to strings gives a compile-time error. However, one could expect GHC to be unhappy about `"4"`, not about `5` as above! Let's take a deeper look on how type checker works.
```
BinderDavid, line 21:
```suggestion
Given that Haskell is a strong statically typed language, it is not a surprise that adding numbers to strings gives a compile-time error. However, one could expect GHC to be unhappy about `"4"`, not about `5` as above! Let's take a deeper look on how the type checker works.
```

message-index/messages/GHC-39999/adding-number-to-string/index.md
```diff
@@ -0,0 +1,25 @@
...
+
+The compiler first encounters the operator `(+)` which has type `Num a => a -> a -> a`. Thus it infers that both arguments of `(+)` should be of the same type `a`. Looking at the first argument `5` does not reveal us what `a` is: it could be `Int`, could be `Double`, etc. But the second argument `4` is certainly `String`, so the compiler concludes that `a ~ String`.
```
BinderDavid, line 23:
```suggestion
The compiler first encounters the operator `(+)` which has type `Num a => a -> a -> a`. Thus it infers that both arguments of `(+)` should be of the same type `a`. Looking at the first argument `5` does not reveal us what `a` is: it could be `Int`, could be `Double`, etc. But the second argument `4` is certainly `String`, so the compiler concludes that `a` should be `String`.
```
The reader of that explanation might not yet have seen type equality constraints, so it might be better to write it out.

BinderDavid (changes requested):
Thanks a lot for the additional example! I have two small nitpicks, if you agree with them we can merge afterwards :)

Bodigrim:
@BinderDavid thanks, updated.

BinderDavid (approved):

PR #490, opened 2024-06-06
Add GHC-17335

hsyl20:


message-index/messages/GHC-17335/dsIncompleteRecSel/before/DsIncompleteRecSel.hs
```diff
@@ -0,0 +1,5 @@
+module DsIncompleteRecSel where
+
+data T = T1 | T2 {x :: Bool}
+
+f a = x a
```
Bodigrim, line 5: Could adding a type signature `f :: T -> Bool` be beneficial for readers?

Bodigrim: I pushed addition of type signatures to the branch.

Bodigrim (approved):

PR #492, opened 2024-06-06
Add GHC-42569

hsyl20:


message-index/messages/GHC-42569/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Standalone kind signatures do not support multiple names
+summary: Standalone kind signatures do not support multiple names
+severity: error
+introduced: 9.6.1
+---
+
+Standalone kind signatures do not support multiple names at the moment.
+See https://gitlab.haskell.org/ghc/ghc/issues/16754 for details.
```
Bodigrim, line 9: I think the description can offer a bit more of exposition and context, like "For terms multiple type signatures can be provided on the same line <here is an example>. However, at the moment (link to GHC issue) such shortcut syntax is not supported for multiple kind signatures on the same line..."

PR #494, opened 2024-06-06
Document GHC-12219: static forms cannot be used in splices

wz1000:


message-index/messages/GHC-12219/index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: static forms cannot be used in splices
```
Bodigrim, line 2: Even while GHC itself starts the error message from lower case, I'd prefer to keep `title`s in title case.
```suggestion
title: Static forms cannot be used in splices
```

message-index/messages/GHC-12219/index.md
```diff
@@ -0,0 +1,15 @@
...
+introduced: 9.6.1
+---
+
+```
+static.hs:7:45: error: [GHC-12219]
+    • static forms cannot be used in splices: static 'a'
+    • In the untyped splice: $(lift (show $ staticKey $ static 'a'))
+```
+
+GHC does not support using a static form in a splice, i.e. obtaining a reference
```
Bodigrim, line 14: I appreciate that it's tedious, but IMHO it's worth to provide readers with more context, especially when it comes to rare features of the language. Those who are already in the know what is a static form and what is a splice are not our target audience. For someone who encountered the error just by making a typo it would help to know which language extensions are involved and have a link to the GHC manual at ready. 

I'd suggest to spell out "TemplateHaskell splices" in full, to make it unambiguous.

PR #498, opened 2024-06-06
Add a `hie.yaml`

georgefst:
This allows us to load all Haskell files in the project using the Haskell Language Server.


Bodigrim (approved):

Bodigrim:
Thanks!

PR #509, opened 2024-06-07
Add GHC-18157

ryndubei:


message-index/messages/GHC-18157/stage-restriction-qq/after/Example.hs
```diff
@@ -0,0 +1,7 @@
+{-# LANGUAGE QuasiQuotes #-}
+module Example where
+
+import QQ (qq)
+
+foo :: String
+foo = [qq|1|]
```
Bodigrim, line 7: I'd imagine it should remain as in "before"?..
```suggestion
foo = [qq|bar|]
```

message-index/messages/GHC-18157/stage-restriction-splice-1/after/Example.hs
```diff
@@ -0,0 +1,8 @@
+{-# LANGUAGE TemplateHaskell #-}
+module Example where
+
+import Foo (foo)
```
Bodigrim, line 4: Could we fix the error with `$(return [])` as described at the end of https://downloads.haskell.org/ghc/9.12.1/docs/users_guide/exts/template_haskell.html#syntax ?

PR #513, opened 2024-06-08
Use haskeline in create script

### LINKED ISSUE #499: Use haskeline for create-message-template.hs
I noticed I cannot use my arrow keys to move the cursor when creating a new message. Delete also doesn't work properly over multiple lines.

The https://hackage.haskell.org/package/haskeline package provides a solution for these problems.

BinderDavid:
Fixes #499 

Uses `haskeline` to allow easier line-editing when using the `create-message-template.hs` script.

### DIFF
--- a/message-index/create-message-template.hs
+++ b/message-index/create-message-template.hs
@@ -15,6 +15,11 @@ import Text.Read (readMaybe)
 
 type ToolM a = InputT IO a
 
+getInputLine' :: String -> ToolM String
+getInputLine' s = do
+  ln <- getInputLine s
+  pure (maybe "" id ln)
+
 -------------------------------------------------------------------------------
 -- Querying the user about the diagnostic
 -------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


BinderDavid:
@noughtmare Can you take a look and verify that it works as intended?

noughtmare:
It might be good to add constraints to the dependencies (at least lower bounds).

message-index/create-message-template.hs
```diff
@@ -141,23 +141,21 @@ validateExampleName "" = False
...
+readExample :: Int -> ToolM String
 readExample i = do
-  putStrLn ""
-  putStrLn ("· Give a name for example " <> show i)
-  putStrLn "The name should not contain spaces and begin with a lowercase letter."
-  putStr "Input: "
-  ln <- getLine
+  outputStrLn ""
+  outputStrLn ("· Give a name for example " <> show i)
+  outputStrLn "The name should not contain spaces and begin with a lowercase letter."
```
jhrcek, line 157: While you're updating this, could you please reword the promt for better clarify?

Yesterday I wasn't sure if it means this:
The name should not ( contain spaces and begin with a lowercase letter ).

or this:

The name should ( not contain spaces ) and (shoud) begin with a lowercase letter.
:smile: 

How about:
```suggestion
  outputStrLn "The name should not contain spaces and should begin with a lowercase letter."
```

jhrcek (approved):
Nice improvement :+1: . Just a small suggestion for one prompt that confused me yesterday.

PR #514, opened 2024-06-08
Add documentation for GHC-21926

FPtje:
This is the card I drew during ZuriHac! :card_index: 

This is quite a simple parser error, which can be hit when `PackageImports` is enabled. The error itself is quite clear already: don't put version numbers and non-alphanumeric characters in a package import.

I'm not sure how one would get into the non-alphanumeric character situation (besides an honest typo), but the version one is quite a fair mistake to make.

### DIFF
--- a/message-index/messages/GHC-21926/index.md
+++ b/message-index/messages/GHC-21926/index.md
@@ -6,6 +6,6 @@ introduced: 9.6.1
 extension: PackageImports
 ---
 
-Version number or non-alphanumeric character in the package. Alphanumeric characters are letters and numbers, as defined by `Data.Char.isAlphaNum`.
+Version number or non-alphanumeric character in the package. Each of dash-separated components of package name must consist of alphanumeric characters (as defined by `Data.Char.isAlphaNum`), at least one of which is not a digit.
 Note that you will also see this error if the package name consists only of digits.
 The package name in this case refers to the one mentioned in an import when using the `PackageImports` language extension.
--- a/message-index/messages/GHC-21926/index.md
+++ b/message-index/messages/GHC-21926/index.md
@@ -5,5 +5,5 @@ severity: error
 introduced: 9.6.1
 ---
 
-Version number or non-alphanumeric character in package name.
+Version number or non-alphanumeric character in package name. Alphanumeric characters are letters and numbers, as defined by `Data.Char.isAlphaNum`.
 The package name in this case refers to the package specified with imports after enabling the `PackageImports` language extension.
--- a/message-index/messages/GHC-21926/index.md
+++ b/message-index/messages/GHC-21926/index.md
@@ -1,0 +1,9 @@
+---
+title: Version number or non-alphanumeric character in package name
+summary: Package imports may only consist of alphanumeric characters, and must omit the version
+severity: error
+introduced: 9.6.1
+---
+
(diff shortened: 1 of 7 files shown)


message-index/messages/GHC-21926/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Version number or non-alphanumeric character in package name
+summary: Package imports may only consist of alphanumeric characters, and must omit the version
+severity: error
+introduced: 9.6.1
+---
+
+Version number or non-alphanumeric character in package name.
```
Bodigrim, line 8: It's probably a good place to describe what's exactly caled "alphanumeric" by GHC.

message-index/messages/GHC-21926/version-number-in-package/index.md
```diff
@@ -0,0 +1,3 @@
+---
+title: Version number in package
```
Bodigrim, line 2:
```suggestion
title: Version number in package name
```

message-index/messages/GHC-21926/index.md
FPtje: Good point! I've tracked down the definition used for the error here:

https://github.com/ghc/ghc/blob/c4a7680afc5a374c668dc80a8abdb3641ab115a2/compiler/GHC/Utils/Misc.hs#L1030-L1033

The error is thrown here:
https://github.com/ghc/ghc/blob/c4a7680afc5a374c668dc80a8abdb3641ab115a2/compiler/GHC/Parser.y#L1155-L1159

This also means that the error is thrown when the package name consists entirely of digits. I added this to the explanation.

message-index/messages/GHC-21926/index.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Version number or non-alphanumeric character in package name
+summary: Package imports may only consist of alphanumeric characters, and must omit the version.
+severity: error
+introduced: 9.6.1
+extension: PackageImports
+---
+
+Version number or non-alphanumeric character in the package. Alphanumeric characters are letters and numbers, as defined by `Data.Char.isAlphaNum`.
```
Bodigrim, line 9: That's a bit incomplete: a package name can include dashes, even while `Data.Char.isAlphaNum '-' = False`. 

The actual function GHC uses to validate package names is 
```haskell
looksLikePackageName = all (all isAlphaNum <&&> not . (all isDigit)) . split '-'
```
So probably something like "All dash-separated components of package name must consist of alphanumeric characters (as defined by `Data.Char.isAlphaNum`), at least one of which is not a digit". What do you think?

FPtje: Sorry for the late reply. Sounds good! Changed the code 👍

Bodigrim: Did you push anything? I don't see any changes.

FPtje: Yikes, rookie mistake. Pushed as [`f1fde5c` (#514)](https://github.com/haskellfoundation/error-message-index/pull/514/commits/f1fde5c6cebe9d594a071d4d00e47c0351c6c5c3).

Apologies for the inconvenience.

Bodigrim (approved):

Bodigrim:
Thanks!

PR #516, opened 2024-06-08
Add index entry for GHC-25955

### LINKED ISSUE #216: PsErrExplicitForall -> 25955

cbrt-x:
Closes #216


message-index/messages/GHC-25955/illegal_explicit_forall/after/Illegal_explicit_forall.hs
```diff
@@ -0,0 +1,7 @@
+{-# LANGUAGE Haskell2010 #-}
+{-# LANGUAGE ExplicitForAll #-}
```
Bodigrim, line 2: This is technically correct, although you almost never want to enable the syntax extension `ExplicitForAll` just for the sake of it. Most likely the very next error message from GHC will complain about missing `ScopedTypeVariables` or `RankNTypes`. This is probably something worth to discuss in `index.md`.

Also worth mentioning that `GHC2021` / `GHC2024` enable this extension by default.

PR #517, opened 2024-06-08
Add index entry for GHC-94803

jllang:


message-index/messages/GHC-67120/index.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Missing main
+summary: Missing main action
+severity: error
+introduced: 9.8.1
+---
+
+According to the second paragraph of Chapter 5 of The Haskell 98 Report,
+>  A Haskell program is a collection of modules, one of which, by convention, must be called Main and must export the value main. The value of the program is the value of the identifier main in module Main, which must be a computation of type IO t for some type t (see Chapter 7). When the program is executed, the computation main is performed, and its result (of type t) is discarded.
```
Bodigrim, line 9:
```suggestion
>  A Haskell program is a collection of modules, one of which, by convention, must be called `Main` and must export the value `main`. The value of the program is the value of the identifier `main` in module `Main`, which must be a computation of type `IO t` for some type `t` (see Chapter 7). When the program is executed, the computation `main` is performed, and its result (of type `t`) is discarded.
```

message-index/messages/GHC-67120/example1/index.md
```diff
@@ -0,0 +1,3 @@
+---
+title: <insert a title for this example here>
```
Bodigrim, line 2: Could you please fill the title?

message-index/messages/GHC-67120/example1/after/Main.hs
```diff
@@ -0,0 +1,6 @@
+module Main where
```
Bodigrim, line 1: My experience is that it's far more common reason for GHC-67120 that you meant to write a library exporting `notMain`, in which case the solution is to rename `module Main` to module `Lib` or anything. 

What contributes to the problem is that GHC would assume `module Main where` implicitly unless there is `module` declaration. 

Probably worth adding another example or at least discussing this case in `index.md`.

PR #519, opened 2024-06-08
Document GHC-10498

Bodigrim:


message-index/messages/GHC-10498/index.md
```diff
@@ -0,0 +1,12 @@
...
+summary: The same entity has more than one definition
+severity: error
+introduced: 9.6.1
+---
+
+GHC does not allow the same entity to have more than one
+definitions. In some contexts it can figure out which one
+supposed to shadow another: local definitions take
+priority over global definitions, etc. Shadowing is just a warning, GHC-63397. However, if the definitions are on the same
+level it becomes a fatal error.
```
jhrcek, line 12:
```suggestion
GHC does not allow the same entity to have more than one
definition. In some contexts it can figure out which one
is supposed to shadow another: local definitions take
priority over global definitions, etc. Shadowing is just a warning, GHC-63397. However, if the definitions are on the same
level it becomes a fatal error.
```

jhrcek (commented):
Looks good, just two small linguistic suggestions

Bodigrim:
@BinderDavid just a gentle ping.

jhrcek (approved):
LGTM

PR #521, opened 2024-06-09
Add error GHC-23482

jaccokrijnen:


message-index/messages/GHC-23482/index.md
```diff
@@ -0,0 +1,8 @@
+---
+title: TypeApplication
+summary: A type application is only allowed with the TypeApplication extension enabled.
+severity: error
+introduced: 9.8.1
+---
+
+Applying a term to a type (using @ notation) requires enabling the TypeApplication extension.
```
Bodigrim, line 8: Could you please quote the text of the error message somewhere? This would help search engines to find this page.
```
/tmp/foo.hs:3:5: error:
    Illegal visible type application ‘@Int’
      Perhaps you intended to use TypeApplications
  |
3 | x = read @Int "42"
  |     ^^^^^^^^^
```

Also it's worth to mention that `TypeApplications` is implied by `GHC2021` / `GHC2024` and with modern GHCs one has to specify `Haskell2010` explicitly to trigger the error.

PR #522, opened 2024-06-09
Add message for GHC-10173

haglobah:


message-index/messages/GHC-10173/index.md
```diff
@@ -0,0 +1,14 @@
+---
+title: No top level binding to annotate
+summary: There is no top level binding this annotation can match.
+severity: error
+introduced: 9.6.1
+---
+
+There is no binding available that this annotation can match.
```
Bodigrim, line 8: Please quote the error message from the compiler here and/or in `index.md` for individual examples, it would help readers to recognise that this is a page they are looking for.

PR #523, opened 2024-06-09
Add error message for GHC-06201

fendor:


message-index/messages/GHC-06201/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Missing instance for required type class method
```
BinderDavid, line 2:
```suggestion
title: Missing method in type class instance
```

Otherwise it might sound like we are missing an instance instead of a method.

BinderDavid (changes requested):
Thanks for the contribution! I have just two small nitpicks that can easily be fixed :)

And sorry that it takes me a while to work on this backlog of new PRs 😅

fendor:
Thanks, incorporated feedback!

BinderDavid (approved):
Thanks. LGTM

PR #524, opened 2024-06-09
Add message for GHC-72771 "Mismatched default method signature"

VeryMilkyJoe:


message-index/messages/GHC-72771/index.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Mismatched default method signature
+summary: The provided default method implementation's type signature does not match the corresponding method type signature.
+severity: error
+introduced: 9.8.1
+---
+
+This error means that there is a type class with some method `m` for which a [default signature](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/default_signatures.html) was defined, but that default signature does not match the type signature of `m`.
```
Bodigrim, line 8: Let's link a stable released documentation instead of GitLab:
```suggestion
This error means that there is a type class with some method `m` for which a [default signature](https://downloads.haskell.org/ghc/9.12.1/docs/users_guide/exts/default_signatures.html) was defined, but that default signature does not match the type signature of `m`.
```

message-index/messages/GHC-72771/example/before/Main.hs
```diff
@@ -0,0 +1,11 @@
...
+module Main where
+import GHC.Generics
+
+main :: IO ()
+main = undefined
+
+class Enum a where
+  enum :: [a]
+  default enum :: (Int ~ a) => a
+  enum = [1..]
```
Bodigrim, line 11: 1. If you rename it to `module Lib where` there would be no need to define `main :: IO ()`, minimizing the example.

2. Let's add `{-# LANGUAGE TypeOperators #-}`, otherwise recent GHCs throw a warning about `~`.

3. Could we please rename `Enum -> Enumerate`, `enum -> enumerate`? So that we do not clash with a usual `class Enum` from `Prelude`.

VeryMilkyJoe: Thank you, that all sounds much better! 
I made changes accordingly.

message-index/messages/GHC-72771/example/after/Lib.hs
```diff
@@ -0,0 +1,11 @@
...
+{-# LANGUAGE TypeOperators #-}
+
+module Lib where
+
+import GHC.Generics
+
+class Enumerate a where
+  enumerate :: [a]
+  default enumerate :: (Int ~ a) => [a]
+  enumerate = [1 ..]
```
Bodigrim, line 11:
```suggestion
  enumerate = [1 ..]
  
```

message-index/messages/GHC-72771/example/before/Lib.hs
```diff
@@ -0,0 +1,11 @@
...
+{-# LANGUAGE TypeOperators #-}
+
+module Lib where
+
+import GHC.Generics
+
+class Enumerate a where
+  enumerate :: [a]
+  default enumerate :: (Int ~ a) => a
+  enumerate = [1 ..]
```
Bodigrim, line 11:
```suggestion
  enumerate = [1 ..]
  
```

Bodigrim (approved):

Bodigrim:
Thanks!

PR #526, opened 2024-06-09
GHC-81995: add error message and example using `sequence`

dschrempf:


message-index/messages/GHC-81995/sequence/before/GetLine.hs
```diff
@@ -0,0 +1,6 @@
+module GetLine where
+
+main :: IO ()
+main = do
+  sequence [putStrLn "First line", putStrLn "Second line"]
```
jllang, line 5: I think we could also have a pure example, e.g., if we define
```
list :: [Int]
list = do
  return 1
  return 2
```
we get `[2]` as the resulting value for `list`.

dschrempf: That's right. I can add another example. I think a classic example using IO is good to have.

dschrempf: @jilang I have added your nice example.

BinderDavid (approved):
Thanks for the contribution! And sorry for the delayed review.

dschrempf:
No problem, and thanks for merging!

PR #529, opened 2024-06-11
Ghc 45102

jappeace:


message-index/messages/GHC-45102/example1/index.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Lambda syntax in `case` expression
```
Bodigrim, line 2: Is it related to GHC-45102? Seems like a rebase artifact.

PR #534, opened 2024-08-04
Basic explanation of 59692

noughtmare:


message-index/messages/GHC-59692/index.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Duplicate instance declarations
+summary: An instance has been defined twice for the same type.
```
ulysses4ever, line 3: It's not necessarily exactly twice, is it? A more precise way to describe this would be, perhaps:

```suggestion
summary: An instance of a class has been defined once again for the same type.
```

ulysses4ever: But for the compiler, there was an instance that it saw in the past, only it's very close past: maybe in the same .hs file (but maybe not). So, I don't see a problem here. But "multiple times" works for me too.

Bodigrim (approved):
It would be nice to have a less artificial example, but it can be left for future work.

Bodigrim:
@noughtmare could you possibly rebase, incorporating already merged #541?

noughtmare:
@Bodigrim done!

Bodigrim (approved):

Bodigrim:
Thanks!

PR #541, opened 2024-10-12
Add documentation and examples for GHC-59692: Duplicate Instances

Qbxs:


message-index/messages/GHC-59692/multipleInstances/index.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Multiple Instances for Semigroup Int
+---
+
+Different definitiones for `Semigroup Int` are possible.
+If we more than one, we ought to wrap `Int` in a `newtype`.
```
BinderDavid, line 6:
```suggestion
If we require both instances, then we should define two `newtype` wrappers for `Int`.
```

message-index/messages/GHC-59692/multipleInstances/index.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Multiple Instances for Semigroup Int
+---
+
+It is not possible to give two definitions for `Semigroup Int`, one for the semigroup given by addition and one for the semigroup given by multiplication.
```
BinderDavid, line 5:
```suggestion
Haskell does not allow to simultaneously define both instances of `Semigroup Int`, one for the semigroup given by addition and one for the semigroup given by multiplication.
```

BinderDavid, line 5:
```suggestion
Haskell does not allow to give more than one definition of `Semigroup Int`, even though we might want to provide both an instance for the semigroup given by multiplication and the semigroup given by addition.
```

BinderDavid (approved):
Thanks! Looks good :)

PR #543, opened 2024-10-13
Document GHC-08838

StillerHarpo:
We need a better example here. It seems to be this warning only occurs with nested returns. I tried
```haskell
f = do
  return (putStrLn "text")
  return ()
```
which didn't produce that warning

### DIFF
--- a/message-index/messages/GHC-08838/WrongDoBind/index.md
+++ b/message-index/messages/GHC-08838/WrongDoBind/index.md
@@ -1,0 +1,16 @@
+
+---
+title: Double return
+---
+
+In this example, there is a nested return whose result is not bound to a variable. Therefore the computation `return 10` is not used and can be removed.
+
+```
+messages/GHC-08838/WrongDoBind/before/WrongDoBind.hs:5:4: warning: [GHC-08838] [-Wwrong-do-bind]
+    A do-notation statement discarded a result of type ‘m Int’
+    Suggested fix:
+      Suppress this warning by saying ‘_ <- return (return 10 :: m Int)’
+  |
+5 |    return (return 10 :: m Int)
+  |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+```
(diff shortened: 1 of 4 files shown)


message-index/messages/GHC-08838/index.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Wrong do bind
+summary: Returning a monadic computation in a do block and not binding it will not run it
+severity: warning
+introduced: 9.4.1
+---
+
+When using a monadic computation in a do block one should either run it 
+directly or bind it. If neither is done the monadic computation is never
```
BinderDavid, line 9:
```suggestion
directly or bind the result of the computation to a variable. If neither is done the monadic computation is never 
```

message-index/messages/GHC-08838/WrongDoBind/index.md
```diff
@@ -0,0 +1,16 @@
+
+---
```
BinderDavid, line 2:
```suggestion
---
```

message-index/messages/GHC-08838/WrongDoBind/index.md
```diff
@@ -0,0 +1,16 @@
...
+title: Double return
+---
+
+In this example, there is a nested return which is not bound. Therefore the computation `return 10` is not used and can be removed.
```
BinderDavid, line 6:
```suggestion
In this example, there is a nested return whose result is not bound to a variable. Therefore the computation `return 10` is not used and can be removed.
```

message-index/messages/GHC-08838/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Wrong do bind
+summary: Returning a monadic computation in a do block and not binding it will not run it
+severity: warning
+introduced: 9.4.1
```
BinderDavid, line 5:
```suggestion
introduced: 9.6.1
```

BinderDavid (approved):
Thanks!

PR #548, opened 2024-10-30
Binding a GADT constructor in let doesn't work

tomsmeding:
People sometimes run into this (e.g. [here](https://ircbrowse.tomsmeding.com/day/lchaskell/2024/10/30?id=1391578#trid1391578)), and the fix is quite non-obvious from the error message. I thought it might be good to expand the pages for the relevant error messages with a relevant example; the existing ones are for very different situations.

I found it very amusing that I needed to edit the pages for _two_ error messages, because the generated error depends on whether `-XTypeFamilies` is active.

The text here is kind of stub; I didn't have time to write an in-depth description. Feel free to improve. But hopefully this is already useful.

### DIFF
--- a/message-index/messages/GHC-25897/example2/after/Let.hs
+++ b/message-index/messages/GHC-25897/example2/after/Let.hs
@@ -1,0 +1,12 @@
+module Main where
+
+data Showable where
+  MkShowable :: Show a => a -> Showable
+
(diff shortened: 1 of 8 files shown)


BinderDavid (approved):
This looks fantastic! Thanks!

PR #549, opened 2024-10-30
Add typed hole example to GHC-88464

tomsmeding:
GHC considers typed holes to be variables that are out of scope, but with special handling in the diagnostics. (Apparently.) So typed holes have the same error number as out-of-scope variables, even though they arise in (from the programmer's perspective) quite different situations.

This adds a simple example and some text explaining how typed holes can be useful.


BinderDavid (approved):
Fantastic text again! Thank you very much.

tomsmeding:
Thanks for maintaining this project! The more of these random bits of wisdom it accumulates, the more useful it gets.

PR #552, opened 2025-01-20
GHC-58481 overloaded labels example

endgame:
The absence of Overloaded Labels can cause confusing parse errors because `#` is lexed as an operator, which then confuses the parser. Try to help the situation by adding an example.

Note: I had to bypass pre-commit hooks when committing the example, because ormolu wanted to format the "before" example and failed because it is designed not to parse. Is there a way to exclude `message-index/messages/GHC-58481/example5/before/Label.hs` from autoformatting?

### DIFF
--- a/message-index/messages/GHC-58481/example5/index.md
+++ b/message-index/messages/GHC-58481/example5/index.md
@@ -1,0 +1,12 @@
+---
+title: Use of overloaded labels without enabled language extension
+---
+
+In this examplem the user attempted to use an overloaded label without enabling the `OverloadedLabels` extension, which leads to this generic parsing error. [Overloaded Labels](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/overloaded_labels.html#extension-OverloadedLabels) assign a special meaning to the `#` character in front of an identifier, or (since GHC 9.6) any non-empty double-quoted string.
+
+```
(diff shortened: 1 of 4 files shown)


message-index/messages/GHC-58481/example5/index.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Use of overloaded labels without enabled language extension
+---
+
+In this examplem the user attempted to use an overloaded label without enabling the `OverloadedLabels` extension, which leads to this generic parsing error. [Overloaded Labels](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/overloaded_labels.html#extension-OverloadedLabels) assign a special meaning to the `#` character in front of an identifier, or (since GHC 9.6) any non-empty double-quoted string.
```
Bodigrim, line 5:
```suggestion
In this example the user attempted to use an overloaded label without enabling the `OverloadedLabels` extension, which leads to this generic parsing error. [Overloaded Labels](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/overloaded_labels.html#extension-OverloadedLabels) assign a special meaning to the `#` character in front of an identifier, or (since GHC 9.6) any non-empty double-quoted string.
```

Bodigrim (approved):

Bodigrim:
Thanks!

PR #555, opened 2025-02-26
Add GHC-67120 `main` not defined

noughtmare:
This is minimum effort, but hopefully still useful.


Bodigrim:
There was a stab at GHC-67120 in #517. Could you possibly incorporate relevant pieces into your branch?

noughtmare:
@Bodigrim I've added two examples, the reference to the Haskell 2010 report, and example error text. I think it is now better than #517.

Bodigrim (approved):

Bodigrim:
Thanks!

PR #576, opened 2026-01-29
Improve documentation for GHC-01239

amalloy:
It talks about using if/then/else as a function argument, but the code snippet just returns it from a function, the same as you could in any language.

I've changed it to use `putStrLn` instead of returning a String.


Bodigrim:
While at it, could we also mention that Haskell `if... then... else...` is what other languages like C call ["ternary conditional operator"](https://en.wikipedia.org/wiki/Ternary_conditional_operator)?

